PR #220, opened 2015-01-08
Allows "symbol" based lookup

andreigabreanu:
Basically if you do

$container->set('something', 'some value');

and then you have

__construct($something) {} 

it will obviously not work. This provides a fix for this case.

### DIFF
--- a/src/DI/Definition/Resolver/ParameterResolver.php
+++ b/src/DI/Definition/Resolver/ParameterResolver.php
@@ -69,6 +69,11 @@ class ParameterResolver
                     continue;
                 }
 
+                if ($this->container->has($parameter->getName())) {
+                    $args[] = $this->container->get($parameter->getName());
+                    continue;
+                }
+
                 throw new DefinitionException(sprintf(
                     "The parameter '%s' of %s has no value defined or guessable",
                     $parameter->getName(),


mnapoli:
Hi, thank you for the contribution!

This feature has been discussed previously, and I am not convinced it is a good idea to introduce it. This would be too much magic and might lead to "wtf" moments where we don't understand what is happening.

andreigabreanu:
Ok. We've rolled out our own implementation anyway thus not using this package so no problem.

FYI though, we've been using "symbol" based lookup for around 2 years and never had a "wtf" moment. The idea to use this approach came from Angular.JS where they inject exclusively based on symbol name (basically a string) and everything is fine overall. 

In ZF2 for example you have the "config" which you can retrieve from the service manager. The config has the value of an array and is not an object.
One example is doing someclass { __construct(array $config) ... } where the $config would be $sm->get('config') -> which returns an array. 
Just some food for thought :)

Cheers!

mnapoli:
Thank you for the feedback.

To be honest this feature is the one I'm most conflicted about and I'm always sad to turn it down. I'm conflicted because of the pros and cons, but also because it would affect greatly how PHP-DI would work and how it would be used, and I don't know if it would go in a better direction.

For example in Angular it's kind of OK to have a service named for example ProductService (even though it shouldn't). But in PHP everything needs to be correctly namespaced and organized (e.g. `Acme\Project\ProductService`). So how would it work? I'm afraid people would tend to resorting to generic/non-namespaced service names, and I'm not sure it would encourage good practices.

Same for your example with `$config`. Having a `config` entry is not really explicit. I agree though that in small apps it's perfectly fine…

So I'm afraid it could be a big can of worms. I understand that it can be useful but I think it comes with a whole thinking on how it would impact best practices and how people configure their apps…

andreigabreanu:
I agree I think with everything you said. In short, I would assume that whoever uses the package, uses it in a responsible manner. I would never do "__construct($productService)" since that's just wrong on so many levels. However, having "a way" to do it if it is the last resort would be somewhat fine I guess. 

Of course i can create a Config class that basically is a wrapper for the config array. But then that's just one more thing to maintain vs. the config array (which even though I hate, that's how the framework imposes it's rules) which doesn't need any configuration.

A newcomer however might have an issue understanding the differences between symbol based and standard class/interface name resolution so indeed maybe it's a bad idea. You could however maybe create a "plugin" for this and allow the dev(s) to integrate it. The idea is that a junior dev would have a hard time integrating the plugin (and why would he devote the time to that when the base package already works).

Or something like that :) 
Anyway, thanks for the discussion!

PR #226, opened 2015-01-27
Simpler closure definitions

mnapoli:
Similarly to [what Pimple did in v2](https://github.com/silexphp/Pimple/pull/83) this PR makes declaring entries using closures simpler.

Before:

```php
return [
    'foo' => DI\factory(function () {
        return new Foo();
    }),
];
```

After:

```php
return [
    'foo' => function () {
        return new Foo();
    },
];
```

`DI\factory()` is now optional, closures will be automatically casted to factory definitions.

**This only works with closures, `DI\factory()` still needs to be used with other callable types**, e.g. invokable object, `['Doctrine\ORM\EntityManager', 'create']`, …

If one wants to inject a callable (i.e. **not** execute it), `DI\value()` needs to be used:

```php
return [
    'foo' => DI\value(function () {
        return new Foo();
    }),
    'bar' => DI\object('stdClass')
        ->constructor(DI\link('foo')),
];
```

This is more complex, but I don't think it's a problem as I've never ever seen that been used (same argument was used in Pimple).

@tsteur, @diosmosis thoughts?


diosmosis:
I think specifying application logic in DI config (ie, injecting closures w/ logic) defeats the purpose of inversion of control, and I think in general defining callback properties is an anti-pattern, so I see no issues.

mnapoli:
@diosmosis we are getting out of topic, but configuration is not application logic, it's configuration. By putting all the configuration and wiring outside of the code, it lets us focus on the actual application logic in the code.

If configuration needs to be PHP code, then so be it.

avant1:
:+1:

tsteur:
:+1:

Issue #208, opened 2014-11-21
Nested definitions


mnapoli:
(splitting issue #190 and merging with #186)

This issue is about supporting nesting definitions, e.g.:

```php
return [
    Db::class => DI\object()
        ->constructor(
            \DI\object(DbConfig::class)
                ->constructor(
                    \DI\env('DATABASE_HOST')
                )
        ),
];
```

mnapoli:
Implemented in 5.0

PR #233, opened 2015-03-06
Add support for nested definitions

mnapoli:
This PR addresses #208 to add support for nested definitions in:

- [x] object definitions (`DI\object()`)
- [x] environment variables (`DI\env()`)
- [x] arrays (`[]` or `DI\add()`)

It also contains a refactoring to simplify the code: `DI\link()` is now considered like any other nested definition (before there was special support for this everywhere because it was the only definition you could use inside another one).

Example:

```php
return [
    Db::class => DI\object()
        ->constructor(
            \DI\object(DbConfig::class)
                ->constructor(
                    \DI\env('DATABASE_HOST')
                )
        ),
];
```


mnapoli:
Merged

PR #236, opened 2015-03-19
Documentation change: Alter factory() example to use safer type-hinting

DHager:
Currently, the example code type-hints with `\DI\Container`, which will break when following the documentation for Symfony-2 integration, since `DI\Bridge\Symfony\SymfonyContainerBridge` is not a subclass.

This problem is avoided by type-hinting with `\Interop\Container\ContainerInterface` instead. (Due to `\DI\ContainerInterface` being deprecated.)

### DIFF
--- a/doc/definition.md
+++ b/doc/definition.md
@@ -210,7 +210,7 @@ return [
     'myNamedInstance' => DI\object('My\Class'),
 
     // Using an anonymous function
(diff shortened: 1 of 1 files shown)


mnapoli:
Yes thank you that's a good change! I've been type-hinting with the ContainerInterface everywhere but too bad it wasn't documented.

PR #244, opened 2015-04-09
Definition extensions

mnapoli:
| | | |
--- | --- | --- | ---
5.0 | [Build Status](https://travis-ci.org/mnapoli/PHP-DI) | [Coverage Status](https://coveralls.io/r/mnapoli/PHP-DI?branch=5.0) | [Scrutinizer Code Quality](https://scrutinizer-ci.com/g/mnapoli/PHP-DI/?branch=5.0)
PR | [Build Status](https://travis-ci.org/mnapoli/PHP-DI) | [Coverage Status](https://coveralls.io/r/mnapoli/PHP-DI?branch=definition-extensions) | [Scrutinizer Code Quality](https://scrutinizer-ci.com/g/mnapoli/PHP-DI/?branch=definition-extensions)

Addresses #214, #234

- refactored definition sources to a simpler design
- class definitions do not extend previous definition anymore #234
- `DI\extend()` allows to create a new "class definition extension" that do extend the previous (or any other) definition
- reworked how the cache is used

Example:

```php
return [
    Foo::class => object(Bar::class)
        ->constructor(...),
];
return [
    // Replace (BC break with PHP-DI 4)
    Foo::class => object(Bim::class)
        ->constructor(...),
    // Extend the previous definition
    Foo::class => extend()
        ->constructorParameter(...),
    // Extend another definition
    Test::class => extend(Foo::class)
        ->constructorParameter(...),
];
```

TODO:

- [x] Implementation
- [ ] Rename `ClassDefinition` to `ObjectDefinition` for consistency
- [ ] `DI\extend()` or `DI\object()->extend()`?
- [ ] Documentation
- [ ] More tests

PR #245, opened 2015-04-11
Definition sources refactoring

mnapoli:
pull request based on #244 but without the feature

This contains only the refactoring of the definition sources to a simpler design + better cache handling.


mnapoli:
The pull request should be against 5.0, not master… trying again

PR #264, opened 2015-06-04
Use $builder instead of $container for useAnnotations and useAutowiri…

holtkamp:
…ng()


mnapoli:
Thank you this was indeed wrong. However I have rewritten (and fixed) this documentation in the `5.0` pull request, e.g. here: https://github.com/mnapoli/PHP-DI/blob/5.0/doc/annotations.md

PR #284, opened 2015-07-24
Improve exception message on unknown method argument

pulyaevskiy:
Hi,

Sometimes when class implementation changes and container definition is out of date it really gets hard to debug where is the issue. Currently exception message contains this:

```
The parameter specified by its name could not be found
```

Also it throws `\ReflectionException` currently. There is nothing in the stack trace which can help in finding invalid definition as well.

This PR improves this situation by:

* Adding name of the argument in the exception message
* Throwing `DefinitionException` which makes more sense since we actually have invalid container definition.

### DIFF
--- a/src/DI/Definition/Helper/ClassDefinitionHelper.php
+++ b/src/DI/Definition/Helper/ClassDefinitionHelper.php
@@ -12,6 +12,7 @@ namespace DI\Definition\Helper;
 use DI\Definition\ClassDefinition;
 use DI\Definition\ClassDefinition\MethodInjection;
 use DI\Definition\ClassDefinition\PropertyInjection;
+use DI\Definition\Exception\DefinitionException;
 use DI\Scope;
 
 /**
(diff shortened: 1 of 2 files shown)


mnapoli:
That looks really good thank you!

mnapoli:
Ah I just realized you made the PR against the `4.x` branch, do you want this released as a new 4.x version? I'll port it to the master branch (5.0) anyway.

pulyaevskiy:
@mnapoli  thanks!

Yes, we still haven't upgraded to 5.0 so having new version tag in 4.x would be very helpful!

pulyaevskiy:
@mnapoli just checking in here: any ETA on new version tag for 4.x?

mnapoli:
Done, sorry for the delay :) https://github.com/PHP-DI/PHP-DI/releases/tag/4.4.9

pulyaevskiy:
no worries! Thanks a lot!

Issue #285, opened 2015-07-24
Support object methods as factories


mnapoli:
```php
return [
    'Foo' => [new FooFactory(), 'create'],
    // these examples are not supported:
    'Foo' => [DI\get('FooFactory'), 'create'],
    'Foo' => ['FooFactory', 'create'],
];
```

Those examples are actually in the documentation, but they do not work. They need to be supported.

For the record they are supported in `$container->call()`.

mnapoli:
Would be covered by #197

mnapoli:
Was closed by #308

PR #295, opened 2015-08-13
Update php-definitions.md

mindplay-dk:
Improve documentation for DI\factory()

### DIFF
--- a/doc/php-definitions.md
+++ b/doc/php-definitions.md
@@ -131,16 +131,6 @@ return [
 ];
 ```
 
-It's usually better to lazy-load `FooFactory` and let the container create it as
-needed - assuming you have auto-wiring enabled, or have already registered a
-definition for `FooFactory`, you can use this definition:
-
-```php
-return [
-    Foo::class => DI\factory([DI\get(FooFactory::class), 'create']),
-];
-```
-
 Or, if `FooFactory::create()` were a static method, you could use this definition:
 
 ```php
@@ -148,6 +138,8 @@ return [
     Foo::class => DI\factory([FooFactory::class, 'create']),
 ];
 ```
+
+(Note that support for lazy creation of a factory instance is planned for a future release.)
 
 When using a system with multiple definition files, you can override a previous entry using a decorator:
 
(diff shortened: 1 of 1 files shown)


doc/php-definitions.md
```diff
@@ -111,24 +111,41 @@ return [
...
+needed - assuming you have auto-wiring enabled, or have already registered a
+definition for `FooFactory`, you can use this definition:
 
+```php
+return [
+    Foo::class => DI\factory([DI\get(FooFactory::class), 'create']),
+];
+```
+
+Or, if `FooFactory::create()` were a static method, you could use this definition:
```
mindplay-dk: Is the static example really necessary or too verbose? Should we drop it?

doc/php-definitions.md
```diff
@@ -111,24 +111,41 @@ return [
...
+];
+```
+
+It's usually better to lazy-load `FooFactory` and let the container create it as
+needed - assuming you have auto-wiring enabled, or have already registered a
+definition for `FooFactory`, you can use this definition:
 
+```php
+return [
+    Foo::class => DI\factory([DI\get(FooFactory::class), 'create']),
```
mindplay-dk: Is this example correct? Will DI\factory() detect and resolve the reference created by `DI\get(FooFactory::class)` inside the callable? (if so, this is not strictly a callable, is it?)

mindplay-dk:
@mnapoli see my comments inline above.

doc/php-definitions.md
mnapoli: Not yet! See #285 (yet another reason to use `Container::call()` to call factories maybe ;)

mnapoli: > (if so, this is not strictly a callable, is it?)

not strictly a callable indeed

doc/php-definitions.md
mnapoli: I think it's the only example we can use for now since #285 isn't implemented.

mindplay-dk: [Removed](https://github.com/mindplay-dk/PHP-DI/commit/671197e3014e13b95df2a4609def037a044643d7) with a reference to #285 for future rollback.

mnapoli:
Perfect thank you!

PR #299, opened 2015-08-14
allow injection of annotation cache

mindplay-dk:


mnapoli:
Thanks for the PR! Since nobody has complained about that before and since then, I'd rather try to keep the API as simple as possible. We'll reopen if it comes up again.

mindplay-dk:
I figured it was good practice for your DI container to actually use DI, but whatever ;-)

Issue #22, opened 2012-11-10
Zend Framework 2 integration


mnapoli:

mnapoli:
Done by @Rastusik, see #301

PR #301, opened 2015-08-20
docs updated for zf2 pridge

Rastusik:
@mnapoli  I've updated the docs for the php di zf2 bridge. Could you review the changes and merge if everything is ok? thanks

### DIFF
--- a/doc/frameworks/zf2.md
+++ b/doc/frameworks/zf2.md
@@ -1,0 +1,136 @@
+---
+layout: documentation
+---
+
+# PHP-DI in Zend Framework 2
+
+## Set up
+
+If you are using ZF2, PHP-DI provides easy and clean integration so that you don't have
+to call the container (thus avoiding the Service Locator pattern).
+
+First, install the bridge to ZF2:
+
+```
+composer require php-di/zf2-bridge
+```
+
+To use PHP-DI in your ZF2 application, you need to edit `application_root/config/module.config.php`:
+
+```php
+    // ...
+    'modules' => [
+        ...
+        'DI\ZendFramework2',
+        ...
+    ],
+    
+    'service_manager' => [
+        // ...
+        'factories' => [
+            'DI\Container' => 'DI\ZendFramework2\Service\DIContainerFactory',
+        ],
+    ],
+```
+
+That's it!
+
+Now you dependencies are injected in your controllers!
+
+As you can see since PHP-DI 5 it's necessary to enable [annotations](../annotations.md) because they are disabled by default.
+
+## Usage
+
+Now you can inject dependencies in your controllers!
+
+For example, here is the GuestbookController of the quickstart:
+
+```php
+class GuestbookController extends AbstractActionController
+{
+    /**
+     * This dependency will be injected by PHP-DI
+     * @Inject
+     * @var \Application\Service\GuestbookService
+     */
+    private $guestbookService;
+
+    public function indexAction()
+    {
+        $this->view->entries = $this->guestbookService->getAllEntries();
+    }
+}
+```
+
+If you'd like to specify the di configuration yourself, create this file: `application_root/config/php-di.config.php`
+and save it with your PHP DI configuration e.g. like this:
+
+```
+return [
+    'Application\Service\GreetingServiceInterface' => Di\object('Application\Service\GreetingService'),
+];
+```
+
+Head over to [PHP-DI's documentation](http://php-di.org/doc/) if needed.
+
+## Fine tuning
+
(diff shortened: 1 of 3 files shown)


doc/frameworks/zf2.md
```diff
@@ -0,0 +1,136 @@
...
+            'DI\Container' => 'DI\ZendFramework2\Service\DIContainerFactory',
+        ],
+    ],
+```
+
+That's it!
+
+Now you dependencies are injected in your controllers!
+
+As you can see since PHP-DI 5 it's necessary to enable [annotations](../annotations.md) because they are disabled by default.
```
mnapoli: I'm not sure I understand what this sentence imply: do user have to do anything? Or are annotations enabled automatically by the ZF2 bridge? Maybe this could be made more clear.

doc/frameworks/zf2.md
```diff
@@ -0,0 +1,136 @@
...
+If you'd like to specify the di configuration yourself, create this file: `application_root/config/php-di.config.php`
+and save it with your PHP DI configuration e.g. like this:
+
+```
+return [
+    'Application\Service\GreetingServiceInterface' => Di\object('Application\Service\GreetingService'),
+];
+```
+
+Head over to [PHP-DI's documentation](http://php-di.org/doc/) if needed.
```
mnapoli: Maybe link to `http://php-di.org/doc/php-definitions.html` instead of the documentation root?

doc/frameworks/zf2.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Enable or disable annotations
+
+```php
+return [
+    'phpdi-zf2' => [
+        'useAnntotations' => true,
+    ]
+];
+```
```
mnapoli: May be good to say if they are enabled by default or not.

doc/frameworks/zf2.md
```diff
@@ -0,0 +1,136 @@
...
+    'phpdi-zf2' => [
+        'cache' => [
+            'namespace' => 'your_di_cache_key',
+            'adapter' => 'redis',
+            'host' => 'localhost', // default is localhost
+            'port' => 6379, // default is 6379
+        ],
+    ]
+];
+```
```
mnapoli: I highly advise against using the Redis cache implementation as it would be much too slow for DI cache. The best cache (and probably the only one that should be used) is the ApcCache, which is fast enough because it's stored locally in memory.

So maybe it's best not to document Redis cache, and document ApcCache instead?

mnapoli:
Hi @Rastusik that's really awesome! I've added a few notes to improve a few things, and then it's good to merge!

mnapoli:
I guess we can close #22 after that :)

doc/frameworks/zf2.md
Rastusik: I just copied this sentence from the zf1 bridg docs. Annotations are disabled by default, there is a config example at the bottom of the page which should be used to enable annotations, I'll create an anchor link for it.

doc/frameworks/zf2.md
Rastusik: I'm not sure about this one, because APC is unstable on php 5.5 and 5.6 regarding this article https://www.devside.net/wamp-server/installing-apc-for-php-5-5 and my personal experiences.

If you prefer the in memory cache (which I also think is the best possible option), I'd recommend the APCu extension mentioned in the link above, since it shouldn't be doing anything that opcache already does and should have only user level cache features implemented.

The cache is not enabled by default and it is currently possible to configure only the two mentioned cache adapters, since I needed to use php-di in the company I work for and I needed file cache for development and redis for production. If there will be a request for enabling the configuration of other cache adapters, I'll gladly implement the features, but I don't want to do something that nobody requires for now.

Redis cache might be a little bit slower, but id didn't occur to me, when annotations were disabled and I'm running a setup of two app servers in two various data centers, which have a shared cache, so redis is a quite reasonable option for that use case, even if it is slower, since APC cache can't be shared between more servers.

Rastusik:
Hi @mnapoli thankc for the input. I have updated the docs file as you requested, except the last issue, which I think is ok as I wrote it, regarding the cirtumstances. What dou you think?

doc/frameworks/zf2.md
mnapoli: Yep that would definitely clear it up!

doc/frameworks/zf2.md
mnapoli: Interesting to know it works well for you in production with Redis. To be honest I only talked about my personal experience where Redis was slower than APCu.

Good for me then!

mnapoli:
All good!

PR #307, opened 2015-09-02
Fixed php-di#306 modifying ParameterResolver and ObjectCreator

bradynpoulsen:
`DI\Definition\Resolver\ParameterResolver` now handles the value by reference

`DI\Definition\Resolver\ObjectCreator` throws a `DI\DependencyException` if an object could not be created by `ReflectionClass` via `newInstanceArgs()` or `newInstance()`

### DIFF
--- a/src/DI/Definition/Resolver/ObjectCreator.php
+++ b/src/DI/Definition/Resolver/ObjectCreator.php
@@ -151,9 +151,9 @@ class ObjectCreator implements DefinitionResolver
             );
 
             if (count($args) > 0) {
-                $object = $classReflection->newInstanceArgs($args);
+                $object = @$classReflection->newInstanceArgs($args);
             } else {
-                $object = $classReflection->newInstance();
+                $object = @$classReflection->newInstance();
             }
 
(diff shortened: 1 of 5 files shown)


mnapoli:
Thanks!

It seems that we shouldn't try to support parameters by reference, the reflection API doesn't support them: see https://3v4l.org/UslNM and https://3v4l.org/i4Sei. Funny that HHVM works fine though.

Anyway, I think we should just keep the "if the object is null, throw an exception" and drop the rest. In theory users should catch warnings, I don't think we should suppress errors with `@` because it could hide other errors happening in the constructors. The exception should still be thrown as a failsafe. Could you also write a test for this? (just a test that checks the exception is thrown)

bradynpoulsen:
In your https://3v4l.org/UslNM example, `ReflectionClass::newInstanceArgs(array $args)` does support pass by reference IF the value in the array is provided to the array by reference, see https://3v4l.org/0E6ST

I agree about suppressing errors though

mnapoli:
Cool I learned something today :)

The updated PR looks awesome, thanks for the tests. merging!

Issue #197, opened 2014-10-26
Use $container->call() for calling DI\factory functions


ewinslow:
Right now when you use DI\factory you have to do:

```php
factory(function(Container $c) {
  return new Thing($c->get(Service::class));
});
```

What I'd really like to do is this:

```php
factory(function(Service $s) {
  return new Thing($s);
});
```

mindplay-dk:
As far as performance, whatever needs to get resolved via argument type-hints is strictly stuff that would need to get resolved by explicit container lookups inside the definition anyhow - likely the performance difference is going to be marginal?

mnapoli:
@mindplay-dk the thing is that it will use reflection on the closure to guess the parameters. Also `$container->call()` goes through the [Invoker](https://github.com/PHP-DI/Invoker) library and I haven't worked on performances on this one (can't guarantee it's the fastest thing ever).

Benchmarks would probably help to get an idea.

Also could you have a look at #239 and let me know if you think it's a good idea? Because it could be useful with this feature (for example for parameters/non-object dependencies).

mindplay-dk:
Okay, our entire framework has been refactored now, so I can profile tomorrow at work for comparison with profiles I ran earlier this week - I have a feeling this is faster than even the lightweight container we had in there before, and the overhead was already negligible, but I will let you know. I can benchmark the total roundtrip time with/without this simple one-line change to see how that performs too. (I'm less interested in raw container performance and more interested in how it performs in real life - even if it was 20% slower but added < 1% total round-trip overhead, for example, I would care more about productivity than bare bones performance of a component that is already "cheap".)

mnapoli:
Great! If you can, use the latest release (5.0.3) as I added a few performance improvements yesterday.

mnapoli:
@mindplay-dk did you have a chance to do performance tests or profiling with this feature (i.e. use `Container::call()` for invoking factory definitions)?

mindplay-dk:
Sorry, no - this may sound odd, after all my excitement about PHP-DI, but we ended up switching to a much simpler container library I had been working on myself, incorporating some of the missing ideas we got from using your library, and some of the absent features we've discussed earlier, including the Angular-style injections.

[...]

Issue #81, opened 2013-06-24
Integration with Doctrine's RepositoryFactory


mnapoli:
See https://github.com/doctrine/doctrine2/pull/695

It seems the feature will only be introduced in Doctrine 2.5 though.

holtkamp:
:+1: 
This would be nice indeed. Some examples for the use with Doctrine Repositories as well

Got PHP-DI working with some ZendFramework 1 controllers. But now I am trying to inject Doctrine Repositories into my Domain Services. 

Because the Repositories need the Doctrine EntityManager during construction, this fails since the EntityManager is 'not instantiable'. 

For example, for my OrderDomainService, which depends on the OrderRepository:

```
Entry ProjectName\Domain\Service\Order cannot be resolved: 
Entry ProjectName\Domain\Repository\Entity\OrderRepository cannot be resolved: 
Entry Doctrine\ORM\EntityManager cannot be resolved:
class Doctrine\ORM\EntityManager is not instantiable

Full definition:
Object (
    class = #NOT INSTANTIABLE# Doctrine\ORM\EntityManager
    scope = singleton
    lazy = false
)
Full definition:
Object 
(
    class = Caos\Domain\Repository\Entity\OrderRepository
    scope = singleton
    lazy = false

    __construct(
        $em = link(Doctrine\ORM\EntityManager)
        $class = link(Doctrine\ORM\Mapping\ClassMetadata)
    )
)
Full definition:
Object (
    class = ProjectName\Domain\Service\Order
    scope = singleton
    lazy = false
    __construct(
        $orderRepository = link(ProjectName\Domain\Repository\Entity\OrderRepository)
    )
)
```

mnapoli:
With the current approach you first need to store the entity manager in PHP-DI. Then you can get repositories from the entity manager:

```php
'Caos\Domain\Repository\Entity\OrderRepository' => DI\factory($c) {
    $em = $c->get('Doctrine\ORM\EntityManager');
    return $em->getRepository('Caos\Domain\Repository\Entity\Order');
},
```

holtkamp:
Ok, thanks, and for the Entity Manager it would be okay to have the ```DI\factory``` fetch it from the Bootstrap?

```
'Doctrine\ORM\EntityManager' => DI\factory(
        function (\DI\Container $c) {
            $bootstrap = Zend_Registry::get('bootstrap');
            $bootstrap->bootstrap('Doctrine\ORM\EntityManager');
[...]

PR #308, opened 2015-09-02
Resolve/instantiate factories using the container

### LINKED ISSUE #179: Support for DI\link() in factory definitions (arrays)
```php
return [
    'Foo' => DI\factory([DI\link('FooFactory'), 'create']),
];
```

mnapoli:
Fix #179, start of #197

```php
return [
    'Foo' => DI\factory(['FooFactory', 'create']),
];
```

With this configuration, `FooFactory` will be created by the container.

Performance tests and profiling showed a performance regression of up to +40% (worst case) to resolve factories (after working on optimizations). I think this is acceptable given it's only a small part of the whole container (the total performance regression should be negligible, especially since 5.1 already brings good improvements to compensate).

TODO:

- [ ] documentation
- [ ] changelog

### DIFF
--- a/change-log.md
+++ b/change-log.md
@@ -1,5 +1,11 @@
 # Change log
 
+## 5.1
+
+Improvements:
+
+- [#308](https://github.com/PHP-DI/PHP-DI/pull/308): Instantiate factories using the container (`DI\factory(['FooFactory', 'create'])`)
+
 ## 5.0
 
 This is the complete change log. You can also read the [migration guide](doc/migration/5.0.md) for upgrading, or [the news article](news/15-php-di-5-0-released.md) for a nicer introduction to this new version.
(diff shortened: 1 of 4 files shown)


holtkamp:
I am monitoring this issue, mainly due to https://github.com/PHP-DI/PHP-DI/issues/81

>If FooFactory::create() is static, no instance will be created (and the method will be called statically).

Ok, but strange thing is that I have been using this for my repositories for a long time already (5.0?)

```php
Project\Domain\Repository\Entity\ProductTypeRepository::class => DI\factory([Project\Domain\Repository\Entity\ProductTypeRepository::class, 'getInstance'])
```

The 'getInstance' function then uses the EntityManager to fetch the Repository for the Entity (which name is derived from the Repository name). 

Is it correct this 'behavior' is not new?

PR #311, opened 2015-09-05
[Docs] Fix typos and improve wording a bit

jdreesen:
Here, `allows` or `permits` is meant, so `lets` is the correct spelling, because `let's` means `let us`.


mnapoli:
:+1: I make that mistake quite often ;)

PR #319, opened 2015-09-06
DotNotation Definition added

jaakon:
Hi,

I've added a new definition for strings: You can access nested array with the "dot" notation like in Laravel (http://laravel.com/docs/5.1/helpers#method-array-get).

I hope it's of any use for you and you add it to the project.

Cheers,
Jaakon


mnapoli:
Hi, thanks for the pull request! That's interesting, I'll have a look after releasing 5.1 in the next days :)

juliangut:
Ir really looks interesting

```php
doSomething($container->get('settings')['package']['settings']['aSetting']);
```

Is an everyday nightmare

mnapoli:
@juliangut do you use that pattern a lot? Is it mostly for settings like in your example? @jaakon I'm interested to learn too about you use cases?

jaakon:
For me it's the same as for @juliangut, mostly configs:

```
	'translator' => [
		'locale' => 'de',
		'path'   => 'resources/lang/',
	],
```
or 
```
	'stash' => [
		'apc' => [
			'ttl'       => 3600,
			'namespace' => md5($_SERVER['HTTP_HOST'])
		],
		'redis' => [
			'servers' => [['server' => '127.0.0.1', 'port' => '6379', 'ttl' => 10]]
		],
		'sqlite' => [
			'path' => 'storage/cache/'
		]
	]
```

juliangut:
I use it quite often but only for configurations

I see a real benefit on including this PR (even in next release although I know its short timed), as long as $dotNotationResolver is checked after $stringResolver and strings prevail over arrays

```php
[
    'foo.bar' => 'foo',
    'foo' => [
        'bar' => 'bar',
       'baz' => 'baz',
    ],
];

$container->get('foo.bar'); // foo
$container->get('foo')['bar']; // bar
$container->get('foo.baz'); // baz
```

Or am I missing something?

mnapoli:
@juliangut I think there is confusion on how this feature works: `$container->get()` isn't changed, so what you describe doesn't work.

What this PR allows is to do this (if I understood the code correctly):

```php
return [
    MyClass::class => DI\object()
        ->constructor(DI\dot('db.connection.host')),
];
```

Where `DI\dot('db.connection.host')` does the same as `DI\get('db')['connection']['host']` but also does error handling and allows to set a default value. Maybe @jaakon can confirm whether it's supposed to be used that way?

Allowing to do `$container->get('foo.bar.baz')` would break BC because today PHP-DI doesn't forbid users to define an entry named `foo.bar.baz` (so a dot notation could override an existing entry). Maybe we could work around that (by checking if an entry exist first) but it might not be simple and efficient, and also it's a little bit magic so it can be confusing (there is already plenty enough of magic in this container, we need to be careful if we want to add more).

jaakon:
@mnapoli That's correct, it just works if you use the dot() function.

juliangut:
You are right, sorry, I let my imagination fly too high for a moment :)

mnapoli:
Thank you for the pull request but it doesn't seem like it would be solving a lot of problems today. I don't want to bloat this library too much. Maybe we'll revisit later.

PR #320, opened 2015-09-11
Delete unused class

jdreesen:
This class became obsolete since we switched to the 'php-di/invoker' library.


mnapoli:
Oh well good catch :)

PR #321, opened 2015-09-12
Support references to arbitrary container entries for factory definitions

### LINKED ISSUE #272: Support 'Class::method` callables
See also https://github.com/PHP-DI/Silex-Bridge/issues/1

PHP-DI supports this syntax: `['Test\Controller', 'test']` but not `'Test\Controller::test'`.

jdreesen:
Currently, references to container entries in factory definitions are only possible, if they reference a class that actually exist with this exact FQN.

This PR enables you to to reference arbitrary container entries in factory definitions (as long as they are callables, of course), for example:
- `factory('my.closure')`
- `factory('my.invocable.object')`
- `factory(['my.class', 'method'])`

Todo:
- [x] write tests
- [x] update documentation

Closes #272

### DIFF
--- a/change-log.md
+++ b/change-log.md
@@ -1,5 +1,15 @@
 # Change log
 
+## 5.2
+
+Improvements:
+
+- [#272](https://github.com/PHP-DI/PHP-DI/issues/272): Support 'Class::method' syntax for callables
+
+Bugfixes:
+
+- [#321](https://github.com/PHP-DI/PHP-DI/pull/321): Allow factory definitions to reference arbitrary container entries as callables
+
 ## 5.1
 
 Read the [news entry](news/16-php-di-5-1-released.md).
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     "require": {
         "php": ">=5.4.0",
         "container-interop/container-interop": "~1.0",
-        "php-di/invoker": "^1.0.1",
+        "php-di/invoker": "^1.1",
         "php-di/phpdoc-reader": "~2.0"
     },
     "require-dev": {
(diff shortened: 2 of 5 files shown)


mnapoli:
:+1: 

However is the change (current state of the PR) necessary? `is_callable('A::b')` is true (see https://3v4l.org/HgXfa)

jdreesen:
Oh, I didn't know that this is possible! You are right, `is_callable('A::b')` is true for valid PHP callables (if class A actually exists).

But this change uses the container for callable resolution, too, which enables you to use any container entry name for the first part and thus allows something like: `factory('foo.bar.baz::create')`

jdreesen:
I just looked at the documentation where it says `Please note: you can set any container entry name in the array, e.g. factory(['foo_bar_baz', 'create']), allowing you to configure foo_bar_baz and its dependencies like any other object.`

This is actually not possible at the moment and would only be possible with the changes in this PR, too.

mnapoli:
Mhh yeah that's true, 2 good points! I can't believe I didn't add tests in #308…

jdreesen:
Added some unit tests that cover factory definitions that reference container entries.

Tests for `Class::method` callables are failing with `dependencies=lowest`, of course... should we raise the composer version for `php-di/invoker` to `^1.1`?

jdreesen:
Changed name and description of this PR to what it actually does.

mnapoli:
> should we raise the composer version for php-di/invoker to ^1.1?

makes sense to me, yep.

jdreesen:
Added some documentation (please tell me if you have suggestions for improvements on this :).

mnapoli:
This is the cleanest pull request ever! Tests, docs, changelog! That rocks :)

PR #322, opened 2015-09-14
Fix return type annotation

jdreesen:


mnapoli:
:+1: I think the comment ("Value that should be injected in the property") dates back from version 2 where PHP-DI was only working with annotations :p

PR #327, opened 2015-09-17
Improved factory definition/resolver tests

jdreesen:
I realized that my tests for the factory resolver/definitions from #321 were not very good and should be improved.

Thus I refactored them so that the unit tests for the `FactoryResolver` only test what this class really does. The tests that utilize the invoker/container were moved to integration tests where they belong, so that the real behaviour is tested.

I also removed an unused data provider from `DecoratorResolverTest`.

### DIFF
--- a/tests/IntegrationTest/Definitions/FactoryDefinitionTest.php
+++ b/tests/IntegrationTest/Definitions/FactoryDefinitionTest.php
@@ -1,0 +1,144 @@
+<?php
+/**
+ * PHP-DI
+ *
+ * @link      http://php-di.org/
+ * @copyright Matthieu Napoli (http://mnapoli.fr/)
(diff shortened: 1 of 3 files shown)


jdreesen:
Oh, noes! One of the older integration test fails because of the changed exception message in `php-di/invoker` `v1.1.1`.

What should we do? Simply adopting the message in the test wouldn't work because then the test will fail on `dependencies=lowest`.

So...
- require `php-di/invoker` `v1.1.1`? 
- remove the (part of) the exception message?
- maybe some fancy regex exception message annotation would work?

mnapoli:
> require php-di/invoker v1.1.1?

Yep that's the simplest solution, that's what I did in #326. By the way I'll be merging #326 I hope there will not be conflicts, let's see. I'll review the PR itself later today.

jdreesen:
Rebased on master.

PR #328, opened 2015-09-19
Some cleanup

jdreesen:
- Fixed some small leftovers from #326
- Deleted 2 unused `use` statements
- Fixed description in phpdoc for `\DI\value()`
- Replaced double quotes with single quotes in strings


mnapoli:
:+1:

I'll enable https://styleci.io and try to configure it to match existing practices.

Issue #239, opened 2015-03-21
Container::call() should allow to define dependencies a la AngularJS


mnapoli:
Just like in AngularJS we can do:

``` js
angular.controller('MyController', ['$scope', 'dep1', 'dep2', function($scope, dep1, dep2) {
  // ...
}]);
```

It would be great to be able to explicitly ask for dependencies:

``` php
$container->call(['db.host', 'db.port', function ($host, $port) {
    // ...
}]);
```

Currently `Container::call()` guesses the objects to inject using type-hints, but this doesn't always work. By using AngularJS's pattern it's already well known and it solves the problem.

That could also be used (if #197 happens) in `DI\factory` too:

``` php
    'db' => DI\factory(['db.host', 'db.port', function ($host, $port) {
        return new Db($host, $port);
    }]);
```

mnapoli:
I'm leaving that feature aside for now as I'm not 100% sure it's the right way. First it could be confusing to add yet another concept that people have to learn. Second, this doesn't allow to define *some* parameters only:

```php
$container->call(['Psr\Log\LoggerInterface', 'foo', function (LoggerInterface $logger, $foo) {
}]);
```

In the example above I have to redefine the `$logger` parameter even though it could be guessed using autowiring…

mindplay-dk:
> this doesn't allow to define *some* parameters only

Well, it could:

```PHP
$container->call(['foo' => 'bar'], function (LoggerInterface $logger, $foo) {});
```

In other words, array values with numeric keys are applied in order, whereas array values with string keys are applied by name.

Putting the function inside the array is confusing btw, in my example I have the parameter array and function as separate arguments - so as for `DI\factory()`:

```PHP
    'db' => DI\factory(['db.host', 'db.port'], function ($host, $port) {
        return new Db($host, $port);
    });
```

This is backwards incompatible though? Since `DI\factory()` would see `['db.host', 'db.port']` as a `callable`, wouldn't it (?)

mnapoli:
Yes it could work indeed, and your solution would be supported out of the box by `Container::call()` so that's nice.

> This is backwards incompatible though? Since DI\factory() would see ['db.host', 'db.port'] as a callable, wouldn't it (?)

It could simply check if there's a closure as a second parameter. In that case the array is an array of parameters.

However I like the idea of embedding the closure because I'm thinking of "defining" a concept of "extended callables".

[...]

PR #333, opened 2015-09-21
Allow injection of any container object as factory parameter via type hinting

jdreesen:
With this PR it's possible to inject any container object into a factory via type hinting (see #197).

This means that instead of fetching the needed objects from the container inside your factory like this
```php
factory(function(ContainerInterface $c) {
  return new Thing($c->get(Service::class));
});
```
you can now let the container inject your dependencies as arguments to your factory directly
```php
factory(function(Service $s) {
  return new Thing($s);
});
```
-
This works with multiple arguments, too, of course. And it's backwards compatible, so if you don't type hint the *first* parameter, an instance of the `ContainerInterface` will be injected (but that's not recommended anyway).

-
There was a discussion about "angular style DI" in #197 and #239 but as I wasn't sure what's the best way to define this type of dependencies this feature was left out here.

-
I did a quick benchmark (my first one using blackfire) with the provided `blackfire run --samples=100 php factory.php` which showed a performance decrease of about 10% but I'm not sure how reliable this is for real world situations. Maybe you have better ways for benchmarking this?

-
Todo:
- [x] documentation
- [x] changelog

### DIFF
--- a/change-log.md
+++ b/change-log.md
@@ -5,6 +5,7 @@
 Improvements:
 
 - [#272](https://github.com/PHP-DI/PHP-DI/issues/272): Support 'Class::method' syntax for callables
+- [#333](https://github.com/PHP-DI/PHP-DI/pull/333): Allow injection of any container object as factory parameter via type hinting
 
 Bugfixes:
 
--- a/doc/php-definitions.md
+++ b/doc/php-definitions.md
@@ -109,17 +109,18 @@ return [
 ];
 ```
 
-The only parameter of a factory is the container (which can be used to retrieve other entries). You are encouraged to type-hint against the interface `Interop\Container\ContainerInterface` instead of the implementation `DI\Container`: that can be necessary in scenarios where you are using multiple containers (for example if using the PHP-DI + Symfony integration).
+Instances of other classes can be injected through parameters via type-hinting (as long as they are registered within the container or autowiring is enabled).
+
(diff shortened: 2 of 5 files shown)


jdreesen:
Added documentation and entry in the changelog.

mnapoli:
Looks absolutely awesome! I need to test regarding performances too (I have a half-done repository with a few performance tests, I need to finish it a push it sometime soon) but if it's only 10% that seems very good considering how useful it will be.

I'll probably can't review and merge this week, but hopefully I'll have time to look at it soon :)

mnapoli:
Hi, just an update: I moved house and don't have internet access yet, it will be like that for another week, sorry for the delay ;)



Le 22 septembre 2015 à 20:46:16, Jacob Dreesen (notifications@github.com) a écrit:

Added documentation and entry in the changelog.

mnapoli:
Getting back to this, sorry for the delay. I tried it out today performance-wise. Form [the very simple benchmark](https://github.com/PHP-DI/PHP-DI/tree/master/tests/PerformanceTest) it shows a 16% slowdown when using `factory()`: https://blackfire.io/profiles/compare/974a4ada-38aa-472f-97bd-bf5c3220c1bf/graph

It's bother me a little bit even though I really want this feature in. Using closures as factories is a main feature, and when you compare it straight to (for example) Pimple it will be much slower. Maybe there is a little bit of room for improvements to find.

One thing I was looking into lately is caching reflection almost entirely. It's a bit of work and I haven't pushed it on GitHub yet, if you are interested to have a look let me know. Also I have no guarantee at all that it will bring improvements.

jdreesen:
I understand your concerns. Your reflection caching work sounds interesting. Maybe we should wait with this PR until it moved further.

mnapoli:
FYI this PR is included in #347, ready to be merged in master for 5.2 (if you have any comment on #347 feel free)

PR #337, opened 2015-10-11
add convenience method to get method injections for single method

supercoffee:
I added a getter to retrieve the MethodInjections for a single method from the ObjectDefinition. After all, this functionality is available for the properties in the ObjectDefinition.


mnapoli:
Hi, thanks for contributing :) I'm not sure what is the goal for this? This method isn't necessary since it's not used.

mnapoli:
Sorry will close as I don't see where it could be useful yet

Issue #269, opened 2015-06-09
Getting requested name when using wildcards


AydinHassan:
Hi, I tried to look through the docs and code about wilcard definitions but couldn't find what I wanted. Is it possible to get the requested service name in a wildcard factory?

```php
return [
    'Vendor\Package\Command\*Command' => factory(function (Container $c, $requestedName) {
        $cmd = explode("\\", $requestedName)[3]; //DeployCommand
        $cmd = substr($cmd, 0, -7); //Deploy
        $commandService = $c->get(sprintf('%s/Service', $cmd));
        return new $requestedName($commandService);
    }),
];
```

Maybe there is a better way to do what I am trying to accomplish?

mnapoli:
Hi! There is indeed no way to do that for now unfortunately. Let's keep that issue open, I'm currently focused on releasing 5.0 (tomorrow) ;)

AydinHassan:
Cool :) I just had a mess about and got it working by changing

`return call_user_func($callable, $this->container);` to `return call_user_func_array($callable, [$this->container, $definition->getName()]);`

in

`\DI\Definition\Resolver\FactoryDefinitionResolver::resolve`

If you wouldn't mind the change I could PR with some tests?

mnapoli:
I think this change is fine for factories used for wildcards definitions, but introducing this might make it incompatible with #197 (when it's implemented). I think ideally the definition name should only be passed for wildcard definitions and keep other factories unchanged (only the container as a parameter).

AydinHassan:
hmm, okay, that makes sense. It doesn't look like there is an easy way to know if the definition was a wildcard one at that point. I will leave this open and you can point me in the direction when you're a bit less busy!

mnapoli:
Will do! An idea I just thought of: add a list of "extra parameters" in [FactoryDefinition](https://github.com/PHP-DI/PHP-DI/blob/master/src/DI/Definition/FactoryDefinition.php):

- by default the array is empty
- for wildcard definitions the [DefinitionArray](https://github.com/PHP-DI/PHP-DI/blob/master/src/DI/Definition/Source/DefinitionArray.php) will populate the array with the real class name

When the factory is called, do something like:

```php
return call_user_func_array($callable, array_merge([$this->container], $definition->getExtraParameters());
```

What do you think?

AydinHassan:
Yeah that works for me! That would be a BC break right?

mnapoli:
I don't think so, where do you see it?

[...]

PR #345, opened 2015-11-13
Allow injection of container and current definition instance as factory parameters

quimcalpe:
This PR supersedes #333 and #269. Actually, It's done on top of #333 to preserve commits.

You can get now the container and current definition injected into factory closures type-hinted like this:

```php
factory(function(ContainerInterface $c, Definition $d) {
    return new Thing($c->get($d->getName()));
});
```

You can mix also any dependency the container can resolve, parameter order is irrelevant with type-hinting:

```php
factory(function(ContainerInterface $c, Service $s, Definition $d) {
    return new Thing($s, $d->getName(), $c->get("whatever"));
});
```

This is also backwards compatible, so if you don't type hint the first parameter, an instance of the ContainerInterface will be injected, if you don't type hint the second parameter, an instance of the current Definition will be also injected, although that's not recommended anyway.

```php
factory(function($c, $d) {
    $c; // => container instance
    $d->getName(); // => requested name, without wildcards
    ....
});
```

### Todo:
- [x] test case for `FactoryParameterResolver`
- [x] update change-log.md


mnapoli:
I finally got to dive into this! This is a very good addition thanks, I only had a few changes to make. Could you review #347 if you have time? The changes I did are detailed in the PR.

PR #347, opened 2015-11-29
Inject dependencies and requested entry name in factory parameters [retake of #345]

mnapoli:
This PR contains #345 and #333.

The differences with #345:

- optimizations: from my profilings we went from a 25% slowdown on factories to 0%, so 5.2 should not be slower than 5.1 (that was the main blocker for me)
- factories can inject a new object (`DI\Factory\RequestedEntry`) to know the requested definition name: that replaces the injection of `Definition` instances. The reason for that is to avoid coupling user factories with definitions which are an implementation detail of PHP-DI. If definitions ever change, it will not break factories.

Example to get the requested entry name:

```php
factory(function(ContainerInterface $c, RequestedEntry $e) {
    return new Thing($c->get($e->getName()));
});
```

Example to inject other services:

```php
factory(function(Logger $logger, Mailer $mailer) {
    return new Thing($logger, $mailer);
});
```

TODO:

- [x] update changelog
- [x] documentation

This is the last feature planned for 5.2, after that it's release time!


quimcalpe:
Nice optimization passing container in constructor and resolving directly, then you can avoid `TypeHintContainerResolver` in resolver chain :+1: 
`RequestedEntry` is also cleaner... so 2 big thumbs up! :+1: :+1:

mnapoli:
Thanks! Time to merge then :)

jdreesen:
Looks very nice :+1:

PR #356, opened 2015-12-27
Self resolving definitions

### LINKED ISSUE #354: Refactor simple resolvers into the definitions
Let's try to move the "resolver" logic into the definitions for those that are very simple (e.g. [value](https://github.com/PHP-DI/PHP-DI/blob/master/src/DI/Definition/Resolver/ValueResolver.php), [alias](https://github.com/PHP-DI/PHP-DI/blob/master/src/DI/Definition/Resolver/AliasResolver.php), [string](https://github.com/PHP-DI/PHP-DI/blob/master/src/DI/Definition/Resolver/StringResolver.php), ...).

That will avoid many classes, let's then see if it improves performances.

mnapoli:
Fixes #354

Move the "resolver" logic into definitions that are very simple (value, alias, string). That lets us remove several classes (simpler codebase), I'll see if it improves performances.

TODO:

- [x] test performances


mnapoli:
I'm seeing literally no gain with the current changes :(

mnapoli:
After testing again (with improved tests) I'm seeing a very small improvement. I'll be merging this PR not for the performance gains but mostly because it simplifies the codebase.

PR #358, opened 2015-12-29
set composer sample doctrine/cache version

drealecs:
change doctrine/cache version in sample composer code to be in sync with what php-di suggests


mnapoli:
:+1:

PR #364, opened 2016-01-12
Fixed type-hint

holtkamp:


mnapoli:
:+1:

holtkamp:
you're fast ;)

holtkamp:
Mm, maybe even too fast... The BaseClass is abstract and can therefore not be instantiated. When looking at the DataProvider function, fetching the entry identified with the BaseClass, WOULD return a SubClass. Sorry... Confused by the naming...

mnapoli:
Haha no problem that's my mistake, given that was a comment in a test it wasn't a big deal so I just pushed merge, I'll revert it ;)

PR #371, opened 2016-01-24
Add link to the ZF2 bridge repo to the docs

jdreesen:
I just noticed that there is no link to the ZF2 bridge repo in the docs. For the other framework bridges there is one, so I added it for ZF2, too.

### DIFF
--- a/doc/frameworks/zf2.md
+++ b/doc/frameworks/zf2.md
@@ -132,3 +132,7 @@ return [
     ]
 ];
 ```
+
+## More
+
+Read more on the [ZF2-Bridge project on Github](https://github.com/PHP-DI/ZF2-Bridge).


jdreesen:
Another thing that caught my eyes is the inconsistent writing of the name "GitHub" in the docs. Sometimes it's spelled with a lower case "h" and sometimes with an upper case "H". I think the latter one is correct as it's written that way by GitHub itself.

If you'd like I'll create a PR for this with consistent use of "GitHub" throughout the docs.

mnapoli:
Thanks!

> If you'd like I'll create a PR for this with consistent use of "GitHub" throughout the docs.

I try to always write "GitHub" now, I learned the correct writing only recently ;) And yeah I'd merge that pull request

jdreesen:
Okay, then I'll create one soon :)

PR #372, opened 2016-01-24
s/Github/GitHub/g

jdreesen:
Well, there were fewer occurances with lower case "h" than I thought, only in the framework integration docs... it was a bit late last night, I think ;)

I wasn't sure whether to replace it in the class name `GithubProfile` in the `doc/container.md`, too. What do you think?


mnapoli:
:+1: as for the class name it's easier to read as it is so I think it's fine to leave it like that.

PR #380, opened 2016-02-09
Fix #379 Regex characters on container keys are escaped

### LINKED ISSUE #379: Circular dependency error when using wildcard
I'm not sure if I'm simply using features improperly but I don't understand why this generates an error.
This is a piece of my definitions file:
```
'Repository.*' => DI\factory(RepositoryFactory::class),
'Service.*.*' => function(DI\Container $c, \DI\Factory\RequestedEntry $entry){
    $serviceNS = "Application\\Service";
    $parts = explode('.',$entry->getName());
    $model = $parts[1];
    $service = $parts[2];

    $repository = $c->get("Repository.{$model}");
    $serviceClass = "{$serviceNS}\\{$model}\\{$service}";
    return new $serviceClass($repository);
},
```
I get a Circular dependency exception when I get to the line `$repository = $c->get("Repository.{$model}");`. Apparently  "Service.\*.\*" is identified as one of the candidates for "Repository.*", causing the exception.
Is this expected behavior? Am I doing something wrong?

HexSource:


mnapoli:
Thank you, do you have time to add tests? Else I'll manage myself.

mnapoli:
I've merged with tests, thanks again!

PR #392, opened 2016-03-11
Remove executable bit

siwinski:


mnapoli:
Good catch thank you!

PR #394, opened 2016-03-15
Fix #393 Added paragraph about annotation case-sensitivity

### LINKED ISSUE #393: @inject does not work on Amazon ECS
Hey,

I have just deployed my Docker based application to Amazon ECS using an ECS optimized AMI, but I had to notice that all those dependencies of my controllers were null which were injected with annotations. After some debugging I managed to find the very issue:

As I was not aware before that annotations are case-sensitive (I guess it's because of case-sensitivity of certain file systems) I wrote `@inject` everywhere instead of `@Inject`. After correcting this problem, everything worked as expected.

I suggest adding some hints about the case-sensitivity of annotations in the proper section of the documentation. I'll open a PR for this soon.

kocsismate:


doc/annotations.md
```diff
@@ -137,3 +137,5 @@ To check the value of this option, you can run the following command:
 ```
 $ php -i | grep "opcache.load_comments"
 ```
+
+Furthermore, you should mind the case-sensitivity of annotations: for example writing `@inject` or `@injectable` (effectively anything else than `@Inject` and `@Injectable`) can lead to strange bugs on certain systems.
```
mnapoli: Sorry in advance to nitpick :) here is a suggestion for a simpler sentence (I try to keep the docs accessible to people that might not be native english speakers)

> Furthermore, please mind that annotations are case-sensitive. You should write `@Inject` and `@Injectable` instead of `@inject` and `@injectable` to avoid bugs on certain systems.

What do you think?

kocsismate: I like this version!

mnapoli:
Thanks!

PR #401, opened 2016-04-05
Replaced wrong semicolon

ziege:


mnapoli:
Good catch thank you!

PR #407, opened 2016-05-11
Add Algolia search

mnapoli:


mnapoli:
It's live! \o/

Try it out: http://php-di.org/

PR #425, opened 2016-07-07
added use statement to zf1 bridge docs

Anyqax:


mnapoli:
Hi, thanks for the pull request but importing the annotation is not necessary, it works without the `use` statement.

Anyqax:
You´re right... a PHPStorm inspection complained about the missing import. That´s why I tried the namespaced annotation `@DI\Annotation\Injectable()`... which didn´t work, so I added the use statement.

mnapoli:
Right, PhpStorm will complain. However it's not mandatory, Doctrine has an option to allow the annotation not namespaced. The choice is yours :)

mnapoli:
(and regarding the docs I think it's better to keep them as simple as possible)

Issue #343, opened 2015-11-05
Autowiring and Annotations do not work for `DI\object()` inside arrays


mopahle:
Example code:

``` php
<?php
require 'vendor/autoload.php';

class A {
  public function __construct(B $b) {}
}

class B {
  /**
   * @Inject
   */
  public function test() {
    echo "Autowiring and Annotations work.";
  }
}

$builder=new DI\ContainerBuilder();
$builder->useAnnotations(true);
$builder->addDefinitions([
  'test'=>[DI\object(A::class)]
]);
$container=$builder->build();

try {
  $container->get('test');
}
catch (Exception $e)
{
  echo $e->getMessage();
}
```

Executing the code above gives the following output: 

```
Error while resolving test[0]. Entry  cannot be resolved: The parameter 'b' of A::__construct has no value defined or guessable
Full definition:
Object (
    class = A
    scope = singleton
    lazy = false
)
```

changing the definitions to `$builder->addDefinitions(['test'=>DI\object(A::class)]);` works as expected.

juliangut:
Is DI\object inside an array a correct usage? I mean you're already using DI\object to create the resolvable

mopahle:
Having only one item in an array may be not really a usecase. But how would you solve something like this without an array:
```php
$builder->addDefinitions([
  'logappenders'=>[
    DI\object(Logappender1::class),
    DI\object(Logappender2::class),
    DI\object(Logappender3::class),
  ]
])
```
and maybe with a second defintion:
```php
$builder->addDefinitions([
    'logappenders'=>DI\add([
        DI\object(Logappender4::class),
    ])
])
```

mnapoli:
That's weird indeed, thanks for the great bug report. The usage seems valid to me and I think I can see why it's not working… After a very quick look I'm not sure what the best fix would be, I'll try to have a look at it soon.

As a workaround (in the meantime):

```php
[
  'test' => [get(A::class)],
  A::class => object(A::class),
];
```

mopahle:
Thanks for the update. However, the workaround does not work for me (still getting the same error).

tuanmh:
I got the same error here. Any plan to fix it?

mnapoli:
I know what causes the bug (ArrayDefinition doesn't trigger autowiring on sub-definitions) but from what I remember it's not an easy one to fix and I don't have a lot of time available to work on it at the moment (publishing the last Slim bridge should have been 2 weeks ago for example).

[...]

PR #426, opened 2016-07-07
Make autowiring and annotations work for `DI\object()` inside arrays

leonbobster:
quick fix for #343

### DIFF
--- a/src/DI/Definition/ArrayDefinition.php
+++ b/src/DI/Definition/ArrayDefinition.php
@@ -3,6 +3,7 @@
 namespace DI\Definition;
 
 use DI\Definition\Helper\DefinitionHelper;
+use DI\Definition\Helper\ObjectDefinitionHelper;
 use DI\Scope;
 
 /**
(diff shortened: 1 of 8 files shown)


mnapoli:
Thanks for the pull request, however it seems the build failed and that no tests are covering the new behavior. Also this is adding specific behavior where it could be made more generic (i.e. not having custom behavior for `ArrayDefinition` in `SourceChain` - maybe by doing what's described in https://github.com/PHP-DI/PHP-DI/issues/343#issuecomment-195239192).

leonbobster:
FYI: I'm not sure the build failed because of those changes. I have the same 4 fails in tests without my fix.

leonbobster:
Do you mean It should be implemented using existing HasSubDefinition? It looks like by the nature this interface implies singular relation. Should it be extended with special methods or implemented special interface for multiple Sub-definitions? Or in some other way ... ?

mnapoli:
Actually I think I was wrong, `HasSubDefinition` is for definitions that extend another one (now I realize the naming of that interface is maybe not perfect). So this pull request is maybe a good solution, I'll have a deeper look. The big thing missing are tests though.

I had a look at the build fail and I think it's just a matter of rebasing your pull request against master.

leonbobster:
I'll add tests

mnapoli:
Thanks for the pull request, closing per https://github.com/PHP-DI/PHP-DI/issues/343#issuecomment-306035969

PR #428, opened 2016-07-09
Add support for defining factory parameters

### LINKED ISSUE #362: Factory Parameters
It would be really nice if I could define specific parameters in my factory.

Take the following example where I want to build an array of delegates for a handler

```php
class SomeHandlerFactory
{
    public function create(SomeHandler $handler, array $delegates)
    {
        $handler->doSomethingCool();
        array_map([$handler, 'pushDelegate'], $delegates);
        return $handler;
    }
}
```

I would like to define the array of delegates to isolate accessing the container directly to my `config.php`
```php
// config.php
return [
    'SomeHandler' => factory(['SomeHandlerFactory', 'create'])->param('delegates', get('handlerDelegates')),
    'handlerDelegates' => function (ContainerInterface $container) {
        $delegates = [];
        if (Config::get('someSetting')) {
            array_push($delegates, $container->get('DelegateA'));
        }
        if (Config::get('awesomeMode')) {
            array_push($delegates, $container->get('ImpeachTheOffice'));
        }
        return $delegates;
    }
];
```

predakanga:
Fixes #362, matches your method name conventions, etc wherever possible

### DIFF
--- a/src/DI/Definition/Helper/FactoryDefinitionHelper.php
+++ b/src/DI/Definition/Helper/FactoryDefinitionHelper.php
@@ -82,7 +82,7 @@ class FactoryDefinitionHelper implements DefinitionHelper
      *
      * @return FactoryDefinitionHelper
      */
-    public function factoryParameter($parameter, $value)
+    public function parameter($parameter, $value)
     {
         $this->parameters[$parameter] = $value;
 
--- a/src/DI/Definition/Helper/FactoryDefinitionHelper.php
+++ b/src/DI/Definition/Helper/FactoryDefinitionHelper.php
@@ -28,6 +28,11 @@ class FactoryDefinitionHelper implements DefinitionHelper
     private $decorate;
 
     /**
+     * @var array
+     */
+    private $parameters = [];
+
+    /**
      * @param callable $factory
      * @param bool     $decorate Is the factory decorating a previous definition?
      */
@@ -58,9 +63,29 @@ class FactoryDefinitionHelper implements DefinitionHelper
     public function getDefinition($entryName)
     {
         if ($this->decorate) {
-            return new DecoratorDefinition($entryName, $this->factory, $this->scope);
+            return new DecoratorDefinition($entryName, $this->factory, $this->scope, $this->parameters);
         }
 
-        return new FactoryDefinition($entryName, $this->factory, $this->scope);
+        return new FactoryDefinition($entryName, $this->factory, $this->scope, $this->parameters);
+    }
+
+    /**
+     * Defines arguments to pass to the factory.
+     *
+     * Because factory methods do not yet support annotations or autowiring, this method
+     * should be used to define all parameters except the ContainerInterface and RequestedEntry.
+     *
+     * Multiple calls can be made to the method to override individual values.
+     *
+     * @param string $parameter Name or index of the parameter for which the value will be given.
+     * @param mixed  $value     Value to give to this parameter.
+     *
+     * @return FactoryDefinitionHelper
+     */
+    public function factoryParameter($parameter, $value)
+    {
+        $this->parameters[$parameter] = $value;
+
+        return $this;
     }
 }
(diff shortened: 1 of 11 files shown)


mnapoli:
Hi! That looks really good! I have a few comments, I'll put them in the diff directly.

src/DI/Definition/Helper/FactoryDefinitionHelper.php
```diff
@@ -58,9 +63,29 @@ public function scope($scope)
...
+     * should be used to define all parameters except the ContainerInterface and RequestedEntry.
+     *
+     * Multiple calls can be made to the method to override individual values.
+     *
+     * @param string $parameter Name or index of the parameter for which the value will be given.
+     * @param mixed  $value     Value to give to this parameter.
+     *
+     * @return FactoryDefinitionHelper
+     */
+    public function factoryParameter($parameter, $value)
```
mnapoli: I think `factory(MyFactory::class)->parameter('foo', 'bar')` would be simpler than `factory(MyFactory::class)->factoryParameter('foo', 'bar')`.

tests/IntegrationTest/CallFunctionTest.php
```diff
@@ -201,7 +201,7 @@ public function test_not_enough_parameters()
 
     /**
      * @expectedException \Invoker\Exception\NotCallableException
-     * @expectedExceptionMessage "foo" is neither a callable nor a valid container entry
+     * @expectedExceptionMessageRegExp /('|")foo('|") is neither a callable nor a valid container entry/
```
mnapoli: Is there a reason those changes are necessary? I think it might be because the quotes have changed between the time you created the branch and when you created the pull request. If you rebase against master it should all be good.

mnapoli:
This is a really great pull request! Could you add a few functional test cases ([here](https://github.com/PHP-DI/PHP-DI/blob/ad0c4e085e6c4d5ffd290b94caf549eadf2f3328/tests/IntegrationTest/Definitions/FactoryDefinitionTest.php#L14-L14))?

mnapoli:
Oops almost forgot: could you update the documentation here too: https://github.com/PHP-DI/PHP-DI/blob/master/doc/php-definitions.md#factories

predakanga:
Hey, thanks for the feedback!

I'll update the PR shortly, though I'm not sure if rebasing will need me to create a new one - I selected to merge into PHP-DI:5.3 instead of master.

tests/IntegrationTest/CallFunctionTest.php
predakanga: I'm not sure if it's an OS issue, but even on master those changes are needed for me.
The actual issue is that var_export() ([here](https://github.com/PHP-DI/Invoker/blob/master/src/Invoker.php#L56-L61)) is returning 'foo' for me on both OS X and Linux, where the test assumes it should return "foo"

I've uploaded my test logs for [Linux (docker)](https://gist.github.com/predakanga/93e3db40517b751407f8ee98e04f27d5) and [OS X](https://gist.github.com/predakanga/5e53a2e032be8d72ba9269acbe663acd) to demonstrate

src/DI/Definition/Helper/FactoryDefinitionHelper.php
predakanga: Happy to change it - I initially wrote it as ->parameter, changed it to ->factoryParameter to match methodParameter and constructorParameter

tests/IntegrationTest/CallFunctionTest.php
mnapoli: Right I was running tests on master locally but it was passing, I had a deeper look and it's actually the latest version of PHP-DI/Invoker that changed the quotes -_-

I fixed it on master: f3acb12a2a4d68d210f5b216188545dde03745fb If you rebase against master it should be good now. Sorry for that :)

mnapoli:
Is there a reason you wanted to merge that into 5.3? To me it's a big feature, it deserves a minor version so it should go in master so that we can release it as 5.4. I'm afraid you'll have to close this one and reopen a new one though, I don't think you can edit the target branch?

I had a look at the latest commits and the docs + new tests are awesome!

predakanga:
There's no real reason to merge it into 5.3, I just made the initial fork from there.

I'm rebasing it now, I'll open a new PR with that.

PR #445, opened 2016-12-13
Add the possibility to inject scalar values with autowiring

Rezouce:
Usage:
```php
<?php
class Foo {
    public function __construct($param)
    {
        echo $param;
    }
}

$container->set(Foo::class . '::param', 'myValue');
$container->get(Foo::class); // print myValue
```

The Autowiring class take a $scalarEntryFormat as parameter which is the format to match for the parameter value. The default value is: className::parameter

### DIFF
--- a/src/DI/Definition/Source/Autowiring.php
+++ b/src/DI/Definition/Source/Autowiring.php
@@ -13,7 +13,6 @@ use DI\Definition\ObjectDefinition\MethodInjection;
  */
 class Autowiring implements DefinitionSource
 {
-
     /** @var string */
     private $scalarEntryFormat;
 


ekkinox:
:+1:

mnapoli:
Hi, thank you for the pull request.

Have you tried this, and if so, why do you not want to use it?

```php
$container->set(Foo::class, DI\object()->constructor('myValue'));

// or
$container->set(Foo::class, DI\object()->constructorParameter('param', 'myValue'));
```

Here is the documentation: http://php-di.org/doc/php-definitions.html#objects

Rezouce:
Our use case is that we've got a config file which define those parameters and this config file is then injected as an ArrayDefinition into the container.
We don't want to tightly couple a config file, which can also be used in another context to the container implementation.

mnapoli:
I'm not sure I understand, here is what I got:

- you have a config file that isn't dependent on PHP-DI (e.g. it's a raw PHP array, or maybe INI, YAML, whatever)
- you want to reuse those values in PHP-DI config

Is that right?

How about to keep loading your config file in PHP-DI, and then reuse those parameters like this:

```php
return [
    Foo::class => DI\object()->constructor(DI\get('config.myValue'))
];
```

Rezouce:
It'd work, but one of our goal is to not have to define each time what to inject into the constructor but let the container resolve it by itself.
Kind of convention over configuration.
That's also why I changed the Autowiring class constructor, to allow to change this convention if needed.

mnapoli:
> but one of our goal is to not have to define each time what to inject into the constructor but let the container resolve it by itself.

I don't understand then. Here you define the parameter for `Foo`:

```php
$container->set(Foo::class . '::param', 'myValue');
```

So it's not automatic, it's the same as what I've showed you above.

Rezouce:
I'm not saying that you have nothing to do for it to work, you'll obviously have to set those values to your container for him to be able to resolve the classes.
But if you have something like this, your configuration is totally independent from your container, contrary to your solution:
```php
<?php
$config = [
    Foo::class . '::param' => 'myValue',
];

$container = $containerBuilder
    ->addDefinitions($config)
    ->build();
```

mnapoli:
Oh OK I think I understand now. Here are the downside I see:

- that would make PHP-DI support yet another configuration format (which is obviously more work to maintain), but it's also more complex for users, it's confusing, the documentation is more convoluted, and the container is slower
- that configuration format is very custom to you, no other container or framework supports it, so I don't see why PHP-DI should adapt to that custom format (and not you adapt your format to PHP-DI)
- despite that format being independent from PHP-DI or any other container/framework (because it's only a simple array), it's structure is specific -> that makes it have a syntax (or convention), which makes it a "custom" format, and that custom format is not compatible with all existing containers/frameworks, so IMO that nullifies the point of the format being independent of PHP-DI. We could write a lot of formats independent of PHP-DI, that doesn't mean they are meaningful and should all be supported

Given all that, at that moment, I don't see an advantage to add support for that format in PHP-DI. Did I miss something?

Rezouce:
I agree that having multiple ways to configure things may be confusing. I'm not sure for the speed cost, I didn't test it but it looks to me that the impact would be too slim to be notice. Also there is still the possibility to use the cache which completely nullify it.

The fact that it's possible to auto resolve a class just by matching the key to a class name is a convention that some containers have implemented, that's exactly the same principle here. True it hasn't been implemented in any major container yet, but there was also a time where class auto resolving was not a thing either.
Moreover it's really easy to do this same implemention on any container which already manage auto resolving of classes. On the other hand it'd be tremendous for someone to reimplement the PHP-DI definitions in another container.

I think having an easy way to be able to configure how to auto resolve our objects, which also stay container agnostic, is a good thing to have, don't you?

mnapoli:
Looking at the few pros and the many cons, I don't think it's a good thing to add it. It would be too much to gain very little since we can already do that today.

And I don't see the gain with the format, it might be PHP-DI independent but since nobody else uses it, nobody plans on using it (AFAIK) and there's nobody trying to make it a standard across containers then it's the same as any other kind of format we could imagine. The only point I can then judge it is "will it improve PHP-DI" and I don't think it does because it would duplicate the existing format. The "container-agnostic" doesn't come into play IMO.

bradynpoulsen:
I definitely recommend `object()` instead and just configure the constructor parameter in PHP-DI 5. As suggested by [the docs](http://php-di.org/doc/definition-overriding.html#objects) and https://github.com/PHP-DI/PHP-DI/pull/445#issuecomment-266810447, this will only override that single constructor parameter from earlier definitions.
```php
$config = [
    Foo::class => DI\object()->constructorParameter('param', 'myValue')
];
```
This new format doesn't make much sense to me. Especially since `Foo::class . '::param'` is screaming "callable for `Foo`'s (static) `param()` method" concerning PHP callable basics. See [Example #1 Type 4](https://php.net/manual/en/language.types.callable.php#language.types.callable.passing)

PR #447, opened 2016-12-28
Require PHP 7

mnapoli:
Require PHP 7 or greater for PHP-DI 6.

Is that too much of a jump?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,25 +3,23 @@ language: php
 sudo: false
 
 php:
-  - 5.5
-  - 5.6
   - 7.0
   - 7.1
   - hhvm
 
 matrix:
   include:
-    - php: 5.5
+    - php: 7.0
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -3,25 +3,22 @@ language: php
 sudo: false
 
 php:
-  - 5.5
-  - 5.6
   - 7.0
   - 7.1
-  - hhvm
```
bradynpoulsen: It might be worth testing against nightly as an allowed failure

mnapoli: :+1: good idea, done

mnapoli:
@bradynpoulsen definitely the goal ;)

mnapoli:
Let's do this!

Pull requests to introduce strict types in some classes (or even everywhere) are welcome, it will be a bit of an effort to do it alone :)

PR #448, opened 2017-01-01
Allow to build the container without parameters

mnapoli:
All parameters in the constructor are now optional. We can build the container like this:

```php
$container = new Container();
```

This is the same as `ContainerBuilder::buildDevContainer()`, but simpler. Also, it's about time we get back that constructor :)

### DIFF
--- a/change-log.md
+++ b/change-log.md
@@ -2,6 +2,10 @@
 
 ## 6.0
 
+Improvements:
+
+- The container can now be built without parameters: `new Container()`
+
 BC breaks:
 
 - The deprecated `DI\link()` helper was removed, used `DI\get()` instead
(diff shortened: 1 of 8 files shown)


avant1:
Now `ContainerBuilder::buildDevContainer()` is just a more complex way to create a container with default settings, right?
Maybe it can be removed?

mnapoli:
@avant1 exactly yes. I was wondering about that too, but `buildDevContainer()` has the advantage of explicitly telling you that the container is not optimized for speed… I'm on the fence, what do you think?

avant1:
Yeah, method name that explicitly tells that this way of creating container is mainly for dev env is much more noticeable than warning in documentation or phpdoc of constructor.

Since "Getting started" docs article for 6.0 will contain only `new Container()` way, probably new users will use it to create container.
Maybe we can keep both ways at least in "Getting started" to emphasize the use case, and use `new Container()` in other places for brevity.

PR #450, opened 2017-01-06
Various optimizations/performance improvements

mnapoli:
Various improvements regarding performances (see each commit for details). Those were profiled using Blackfire.

Numbers below represent improvements in very specific benchmarks, time actually gained in real applications is much lower.

[kocsismate benchmark](https://rawgit.com/kocsismate/php-di-container-benchmarks/master/var/benchmark.html):

- 15% improvement for test suite 1: [see the profiling](https://blackfire.io/profiles/compare/f1fb6b1f-c601-4fa8-be99-abf40c3ba568/graph)
- 6.7% improvement for test suite 2: [see the profiling](https://blackfire.io/profiles/compare/93c3c644-663b-4b4a-8a26-37f4584cfe46/graph)
- 24% improvement for test suite 3: [see the profiling](https://blackfire.io/profiles/compare/0f8b2628-3f74-463c-9bd2-4aed333d3eef...adda1d9d-881a-4bbb-a0e6-61cf8cc08ae5/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=landscape&settings%5BtabPane%5D=nodes&selected=&callname=main())

Internal performance test:

- 2.3% improvement for test `get.php`: [see the profiling](https://blackfire.io/profiles/compare/ceba1237-4d0a-4ecb-85a4-fd6a8dc49a87/graph)
- 0% difference for test `get-object.php`: [see the profiling](https://blackfire.io/profiles/compare/ab053cdc-7db3-4fef-9331-9bd6fd519377/graph)
- 6.2% improvement for test `call.php`: [see the profiling](https://blackfire.io/profiles/compare/d36561e1-7b7a-4c94-9532-370d09d5ecf7/graph)


src/DI/Container.php
```diff
@@ -107,15 +107,8 @@ public function __construct(
      */
     public function get($name)
     {
-        if (! is_string($name)) {
-            throw new InvalidArgumentException(sprintf(
-                'The name parameter must be of type string, %s given',
-                is_object($name) ? get_class($name) : gettype($name)
-            ));
-        }
-
```
mnapoli: PHP-DI 6 will be for PHP 7 so the strict type will be enforced, we can afford removing that.

src/DI/Definition/Resolver/ObjectCreator.php
```diff
@@ -134,10 +146,26 @@ private function createInstance(ObjectDefinition $definition, array $parameters)
                 $parameters
             );
 
-            if (count($args) > 0) {
-                $object = $classReflection->newInstanceArgs($args);
-            } else {
-                $object = new $classname;
+            // Optimization trick
+            switch (count($args)) {
```
jdreesen: What about argument unpacking via `...` instead of the switch/case?
```php
$object = new $classname(...$args);
```
Did you try it?

mnapoli: Right! I merged the PHP 7 pull request just after doing my benchmarks, that's dumb :) So currently I don't use PHP 7 features, but I should try that!

mnapoli: And now that I think about it it could also be used to call methods/setters too

src/DI/Container.php
CarwynNelson: Maybe I'm missing something but the `get` method doesn't have a type hint yet?

mnapoli: @CarwynNelson not yet indeed, but I just merged #447 so it will be added in the 6.0 branch before it's released. Removing that check yield a good improvement (because this method is called a lot).

By the way even if we don't add the strict type, it's interesting to note that [Zend Framework's ServiceManager](https://github.com/zendframework/zend-servicemanager/blob/master/src/ServiceManager.php#L176-L184) doesn't have that check (which explains why it's faster).

src/DI/Definition/Resolver/ObjectCreator.php
mnapoli: Oh well, it crashes Blackfire (or PHP 7.1… segmentation fault) :/

Let's try that again in a month or two

kocsismate:
I have great news @mnapoli ! I polished my benchmark in the last couple of days so now results are much more reliable than before. That's why I also wanted to compare PHP-DI 5.4 to 6:

- here are the results of v5.4: https://rawgit.com/kocsismate/php-di-container-benchmarks/c6e2ab7027069d35d848a4e339352a9c3e8b9f1a/var/benchmark.html

- and here is for v6: https://rawgit.com/kocsismate/php-di-container-benchmarks/master/var/benchmark.html

This is a solid 5-10% improvement in all tests! Congratulations :)

mnapoli:
@kocsismate ha I've seen the commit before seeing this comment 😄  Thanks for the update, and thanks for doing that!

PR #456, opened 2017-03-05
Delete .phpstorm.meta.php

zored:
I don't think it's the best solution but this file conflicts with new folder format of [`.phpstorm.meta.php`](https://confluence.jetbrains.com/display/PhpStorm/PhpStorm+Advanced+Metadata). If you will create project, require PHP-DI and create `.phpstorm.meta.php\sample.meta.php`, then autocomplete in PhpStorm will be broken.


zored:
Just was trying to add PHP-DI support in my new big project, but it ruins PhpStorm because of this file. Any suggestions? :)

avant1:
@zored take a look: http://php-di.org/doc/ide-integration.html#phpstorm-plugin

It works fine for me.

mnapoli:
@zored how are you installing PHP-DI? With Composer?

Because this file exists for autocompletion in the PHP-DI project itself, not in projects that use PHP-DI. I would be expecting PhpStorm to ignore `.phpstorm.meta.php` files that are in sub-directories in `vendor/`

Also are you on a EAP?

If anyone else can reproduce/not reproduce please speak :)

zored:
@avant1 Oh, thanks. Great plugin.
@mnapoli I've got PhpStorm 2016.3.2. I've got some observations about this file. [Here is repo](https://github.com/zored/phpstorm-meta). So multiple `.phpstorm.meta.php` files will extend each other. But here is the problem: when I use new format at root folder and old format at subfolder then autocompletions stops working at all. I suppose that it's PhpStorm bug that isn't easy to solve here. So I will now use old format in my project and create issue for JetBrains. 😃

zored:
Here is PhpStorm issue: https://youtrack.jetbrains.com/issue/WI-35301

mnapoli:
Thanks for the info!

PR #467, opened 2017-04-10
Register the container against the PSR ContainerInterface.

rayward:
The container previously used to register itself against `DI\ContainerInterface` (which was subsequently removed). 

Now that PSR-11 has been accepted, it would be useful to be able retrieve the container from the PSR `ContainerInterface`.


mnapoli:
Thanks!

PR #469, opened 2017-04-17
Replace Doctrine Cache with PSR-16

### LINKED ISSUE #376: Support PSR-6 or PSR-16 caching
PHP-DI should support PSR-6 compliant caching instead of just Doctrine's cache implementation. This will ease integration in existing cache setups in the future.

_This is even more important since Symfony 3.1 will ship with a PSR-6 compliant cache component._

Further reading:
- [PSR-6 specification](http://www.php-fig.org/psr/psr-6/)
- [Symfony PSR-6 implementation pull-request](https://github.com/symfony/symfony/pull/17408)

mnapoli:
Fix #376 

The cache system has been moved from Doctrine Cache to PSR-16 (the simple cache standard).

Symfony Cache is recommended. Only the 3.3 version will support PSR-16, and it is not released yet. However 3.3 will be released in May and I think PHP-DI 6 will be released at the same time so it's fine IMO. Also it's a standard, people can use any implementation.

TODO:

- [x] Migration guide

### DIFF
--- a/src/DI/Cache/ArrayCache.php
+++ b/src/DI/Cache/ArrayCache.php
@@ -2,76 +2,90 @@
 
 namespace DI\Cache;
 
-use Doctrine\Common\Cache\Cache;
-use Doctrine\Common\Cache\ClearableCache;
-use Doctrine\Common\Cache\FlushableCache;
+use Psr\SimpleCache\CacheInterface;
 
 /**
  * Simple implementation of a cache based on an array.
  *
- * This implementation can be used instead of Doctrine's ArrayCache for
- * better performances (because simpler implementation).
+ * The code is based on Doctrine's ArrayCache provider.
  *
- * The code is based on Doctrine's ArrayCache provider:
- * @see \Doctrine\Common\Cache\ArrayCache
- * @link   www.doctrine-project.org
+ * @link www.doctrine-project.org
  * @author Benjamin Eberlei <kontakt@beberlei.de>
  * @author Guilherme Blanco <guilhermeblanco@hotmail.com>
  * @author Jonathan Wage <jonwage@gmail.com>
  * @author Roman Borschel <roman@code-factory.org>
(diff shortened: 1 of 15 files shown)


sagikazarmark:
I would probably mention [PHP Cache](http://www.php-cache.com) as well.

src/DI/Cache/ArrayCache.php
```diff
@@ -2,76 +2,90 @@
...
+ * @link www.doctrine-project.org
  * @author Benjamin Eberlei <kontakt@beberlei.de>
  * @author Guilherme Blanco <guilhermeblanco@hotmail.com>
  * @author Jonathan Wage <jonwage@gmail.com>
  * @author Roman Borschel <roman@code-factory.org>
  * @author David Abdemoulaie <dave@hobodave.com>
+ * @author Matthieu Napoli <matthieu@mnapoli.fr>
  */
-class ArrayCache implements Cache, FlushableCache, ClearableCache
+class ArrayCache implements CacheInterface
```
sagikazarmark: Might make sense to have this implementation in a PSR cache util package?

mnapoli:
Good point about PHP cache, I'll add a note in the docs!

src/DI/Cache/ArrayCache.php
mnapoli: This implementation has 2 goals:

- provide a very simple implementation without having to install a separate package
- provide a very fast implementation for benchmarks

The second point is almost stupid and out of scope for end users, but benchmarks are taken too seriously by too many people and Doctrine's implementation was unnecessarily "slow". After giving it a look, I guess Symfony too. When you remove namespaces and that kind of small features, that makes a difference in micro-benchmarks.

mnapoli: Right I see now sorry ;)

PR #471, opened 2017-04-18
Return proper type AutowireDefinitionHelper in PHPDoc

holtkamp:
Maybe even consider `@return $this`?

Currently PHPStorm does not 'accept' this:
- `autoware()->lazy()->constructorParameter()`. 

This is because `lazy()` PHPDoc suggests a CreateDefinitionHelper is returned which does not have method `constructorParameter()`...

screen shot 2017-04-18 at 20 09 40

### DIFF
--- a/src/DI/Definition/Helper/AutowireDefinitionHelper.php
+++ b/src/DI/Definition/Helper/AutowireDefinitionHelper.php
@@ -24,7 +24,7 @@ class AutowireDefinitionHelper extends CreateDefinitionHelper
      * @param string $parameter Parameter for which the value will be given.
      * @param mixed  $value     Value to give to this parameter.
      *
-     * @return CreateDefinitionHelper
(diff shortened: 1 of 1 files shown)


mnapoli:
Does `@return $this` work in PhpStorm? If it does then 👍  to do that, should I wait before merging that PR?

holtkamp:
Yes, `@return $this` works smoothly. 

screen shot 2017-04-19 at 10 32 29

Since that is a bigger change / PR to do that in the entire codebase I thought: this might deserve a separate PR / some discussion?

mnapoli:
:+1: I'm merging this one for now, and if you have time to update the phpdoc that would be awesome.

> Since that is a bigger change / PR to do that in the entire codebase

I think it is only useful for "definition helpers"? (i.e. not the whole codebase)

PR #474, opened 2017-04-19
PSR16 cache key char restriction

### LINKED ISSUE #473: cache key character request
This is a humble question and request

Testing new psr16 in dev-master and using `doctrine/cache` with `php-cache/doctrine-adapter` I get this error from php-cache:

```
1) /.../ContainerBuilderTest::testCreation
Cache\Adapter\Common\Exception\InvalidArgumentException: Invalid key: "DI\Definition\foo". The key contains one or more characters reserved for future extension: {}()/\@:

/.../vendor/cache/adapter-common/AbstractCachePool.php:290
/.../vendor/cache/adapter-common/AbstractCachePool.php:122
/.../vendor/cache/adapter-common/AbstractCachePool.php:428
/.../vendor/php-di/php-di/src/DI/Definition/Source/CachedDefinitionSource.php:73
/.../vendor/php-di/php-di/src/DI/Definition/Source/CachedDefinitionSource.php:44
/.../vendor/php-di/php-di/src/DI/Container.php:194
/.../tests/PHPDI/ContainerBuilderTest.php:78
```

This is the line that triggers the error:

```php
self::assertTrue($container->has('foo'));
```

I know it's no PHP-DI error but some sort of characters php-cache reserved for "the future"

BUT

@mnapoli could you consider using a character other than back-slash given it could have special meaning? underscore maybe?

juliangut:
Changed back slash to underscore to comply with PSR16 cache key restrictions

> Key - A string of at least one character that uniquely identifies a cached item. Implementing libraries MUST support keys consisting of the characters A-Z, a-z, 0-9, _, and . in any order in UTF-8 encoding and a length of up to 64 characters. Implementing libraries MAY support additional characters and encodings or longer lengths, but must support at least that minimum. Libraries are responsible for their own escaping of key strings as appropriate, but MUST be able to return the original unmodified key string. The following characters are reserved for future extensions and MUST NOT be supported by implementing libraries: {}()/\\@:

Fixes #473

### DIFF
--- a/src/DI/Definition/Source/CachedDefinitionSource.php
+++ b/src/DI/Definition/Source/CachedDefinitionSource.php
@@ -17,7 +17,7 @@ class CachedDefinitionSource implements DefinitionSource
      * Prefix for cache key, to avoid conflicts with other systems using the same cache.
      * @var string
      */
-    const CACHE_PREFIX = 'DI\\Definition\\';
+    const CACHE_PREFIX = 'DI_Definition_';
 
     /**
      * @var DefinitionSource


mnapoli:
Thanks!

PR #475, opened 2017-04-19
Use @return $this in PHPDoc to allow fluent method calls in IDE’s

holtkamp:
Also discussed in https://github.com/PHP-DI/PHP-DI/pull/471


mnapoli:
Thanks!

PR #481, opened 2017-04-19
Enforce stricter typing with PHP 7

mnapoli:
Strict typing everywhere!

### DIFF
--- a/src/DI/Annotation/Inject.php
+++ b/src/DI/Annotation/Inject.php
@@ -31,7 +31,6 @@ final class Inject
     private $parameters = [];
 
     /**
-     * @param array $values
      * @throws AnnotationException
      */
     public function __construct(array $values)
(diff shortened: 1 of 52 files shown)


holtkamp:
@mnapoli did you consider bumping the minimum required version to PHP 7.1 to allow nullable types?

This would enable to increase strictness even further, [for example here](https://github.com/PHP-DI/PHP-DI/blob/9f8c660a0a824531c1242ee8b8aedfb0303adf6f/src/DI/Annotation/Injectable.php#L60):

```php
    /**
     * @return bool|null
     */
    public function isLazy()
    {
        return $this->lazy;
    }
```

```php
    /**
     * @return bool|null
     */
    public function isLazy(): ?bool
    {
        return $this->lazy;
    }
```

mnapoli:
@holtkamp, @Taluu, @avant1 I've opened #483 to discuss that.

PR #482, opened 2017-04-19
FQN are no good as cache keys for PSR-16

juliangut:
Following on #474 

When using FQN as definitions these are conflicting with PSR-16 cache key restriction:

```php
$containerBuilder->setDefinitionCache(new PSR_16_cache_implementation());
$containerBuilder->addDefinitions([
    LoggerInterface::class => function (): LoggerInterface {
        return (new Logger())
            ->pushProcessor(new MemoryUsageProcessor())
            ->pushProcessor(new MemoryPeakUsageProcessor())
            ->pushHandler(new ErrorLogHandler(ErrorLogHandler::OPERATING_SYSTEM, Logger::DEBUG));
    },
]);

$container->get(LoggerInterface::class);
```

Produces the error

> Invalid key: "DI_Definition_Psr\Log\LoggerInterface". The key contains one or more characters reserved for future extension: {}()/\\@:

In order to fix this cache key is constructed by changing back slashes to dots, but perhaps dot is not the proper character to use as separator and a less common char should be used instead, such like #, %, * or ~

### DIFF
--- a/src/DI/Definition/Source/CachedDefinitionSource.php
+++ b/src/DI/Definition/Source/CachedDefinitionSource.php
@@ -93,6 +93,6 @@ class CachedDefinitionSource implements DefinitionSource
      */
     private function getCacheKey(string $name) : string
     {
-        return self::CACHE_PREFIX . str_replace('\\', '.', $name);
+        return self::CACHE_PREFIX . str_replace(['{', '}', '(', ')', '/', '\\', '@', ':'], '.', $name);
     }
 }
--- a/src/DI/Definition/Source/CachedDefinitionSource.php
+++ b/src/DI/Definition/Source/CachedDefinitionSource.php
@@ -17,7 +17,7 @@ class CachedDefinitionSource implements DefinitionSource
      * Prefix for cache key, to avoid conflicts with other systems using the same cache.
      * @var string
      */
-    const CACHE_PREFIX = 'DI_Definition_';
+    const CACHE_PREFIX = 'DI.Definition.';
 
     /**
      * @var DefinitionSource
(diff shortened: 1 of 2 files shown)


src/DI/Definition/Source/CachedDefinitionSource.php
```diff
@@ -83,8 +81,18 @@ private function fetchFromCache(string $name)
...
+    /**
+     * Get normalized cache key.
+     *
+     * @param string $name
+     *
+     * @return string
+     */
+    private function getCacheKey(string $name) : string
+    {
+        return self::CACHE_PREFIX . str_replace('\\', '.', $name);
```
Taluu: what about a sha1 key or something like that ?

mnapoli: We'll want that code to be as fast as possible. I even wonder if we can't find a faster solution?

mnapoli:
Well spotted, would you have time to also add tests to ensure there is no regression in the future?

src/DI/Definition/Source/CachedDefinitionSource.php
juliangut: I'd like to test if slashes are really part of a FQN cause just transforming them directly might be hiding user mistakes when naming definitions.

Something like `class_exists($name) || interface_exists($name)` would do, but it would slow down execution

juliangut: Forget what I've said about definition names and caching. The end user shouldn't bother about how PHP-DI caches defintions so I'm going to directly replace all the invalid chars (`{}()/\\@:`) into dots and update tests and this should be done

mnapoli:
Awesome thanks!

PR #489, opened 2017-05-22
Update autowiring.md (link to PHP manual)

Muukuro:
The external link about type hinting had its resource moved to the proposed new URL.


mnapoli:
Thanks!

PR #490, opened 2017-05-22
[WIP] Recursively resolve array definitions

yuloh:
Hello,

I was using PHP-DI and noticed that array definitions are only resolved at the first level of the array.  For example:

```php
return [
    's3.config' => [
        'version'                  => 'latest',
        'region'                   => DI\env('S3_REGION'), // this resolves
        'bucket'                   => 'stuff',
        'credentials'              => [
            'key'    => DI\env('S3_KEY'),
            'secret' => DI\env('S3_SECRET'),  // these don't
        ],
    ],
    // ...
];
```

To make this work I changed the ArrayResolver to recurse and added a test.

It looks like `ArrayDefinition::__toString` would need to change too if this added.  I'm not too sure of what the format should be.

### DIFF
--- a/src/DI/Definition/Resolver/ArrayResolver.php
+++ b/src/DI/Definition/Resolver/ArrayResolver.php
@@ -43,11 +43,11 @@ class ArrayResolver implements DefinitionResolver
         $values = $definition->getValues();
 
         // Resolve nested definitions
-        foreach ($values as $key => $value) {
+        array_walk_recursive($values, function (&$value, $key) use ($definition) {
             if ($value instanceof DefinitionHelper) {
-                $values[$key] = $this->resolveDefinition($value, $definition, $key);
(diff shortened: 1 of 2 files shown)


mnapoli:
Thanks, the reason it wasn't done before is because of the possible performance impact. I think I will need to measure this (I have benchmarks in the repository).

mnapoli:
A quick update: I think the subject of performances may be gone if I succeed with #494. So I will be keeping this pull request around until I can adapt that to the new architecture. I'll let you know, it might take me some time to complete #494

mnapoli:
Thanks! I've merged manually and apply the same behavior to the compiled container.

I've also opened #498 to apply that kind of logic everywhere.

yuloh:
Awesome, thanks!

PR #491, opened 2017-05-24
Cache refactoring for performance improvements

### LINKED ISSUE #477: Refactor the cache to cache all definitions at once?
I think the main way to achieve a huge performance improvement would be to change the way definitions are cached.

Currently each definition is cached under its own key. So on each request, each service entry means one cache call. That's why APCu is recommended and Redis is not (because that would mean hundred of cache calls through the network).

I think the whole cache system can be refactored into something better:

- currently we cannot cache *all* definitions under one cache entry because:
    - we don't know all the possible container entries because of autowiring
    - if we had one cache entry and updated it each time we cache a new definition, we would run into race conditions
- the cache cannot be warmed up

I think we could have a middle ground:

- the warmup could take all definitions **explicitly defined in the config**, resolve them and cache them
- when resolving an entry, the container will then hit the cache for entries in the config, and for new entries that are not in the config (autowiring/annotations) those can get a separate cache key

That means that we could save a good amount of cache queries. And that way also, users that want to optimize for maximum performances could add `foo::class => autowire()` configuration for each class (just to benefit from the warmup/grouped cache).

One thing that will not be easy to deal with is wildcard :) But in a first version maybe those can be ignored (performances will not be worse anyway).

mnapoli:
🎉   yuuuuge performance improvements incoming :)

Fixes #477 This pull request refactors the whole cache system.

## Changes

Before:

- every (cacheable) definition is cached as a separate entry
- each definition is `get` on demand
- each definition is `set` when not found
- all PSR-16 caches are supported
- APCu is recommended because given the number of get/set to the cache, all other caches are too slow

After:

- only APCu is supported: that allows a much more optimized code, less errors from users, and a more "plug and play" solution (just enable the cache, nothing to configure)
- all definitions are cached under a single cache entry
- all the definitions are `get` at once: huge gain
- each definition is `set` when not found (not optimisation here)

Things to look out for: cache stampede.

### Benchmark `get-cache.php`

Before:

- no cache used: 224ms
- APCu cache used: 181ms
- gain thanks to the cache: **19%**
- [profiling](https://blackfire.io/profiles/compare/702a492a-f462-4aed-b210-8ab7ecf2d1eb/graph)

After:

- no cache used: 225ms
- APCu cache used: 146ms
- gain thanks to the cache: **35%**
- [profiling](https://blackfire.io/profiles/compare/29367b6a-b2c5-4e4b-8d98-aec20a1c8feb/graph)

When you use the cache there is a **19% improvement** (181 -> 146ms). [Profiling](https://blackfire.io/profiles/compare/cf181d0a-7629-4d21-aee6-ed78550573f4/graph)

## TODO

- [x] POC
- [x] ~~optimize APCu calls to avoid many `set` in the same `Container::get()` call~~ Not interesting
- [ ] test with 3rd party DI container benchmarks
- [ ] update the `ContainerBuilder`
- [ ] configurable cache key prefix
- [ ] the cache key should contain a hash of all sources? auto-refresh?
- [ ] decide what to do with `ArrayCache`
- [ ] update documentation
- [ ] changelog

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ matrix:
       env: dependencies=lowest
 
 before_script:
+  - phpenv config-add travis.php.ini
   - if [[ $(phpenv version-name) == '7.1' ]]; then composer require satooshi/php-coveralls:dev-master -n ; fi
   - if [[ $(phpenv version-name) != '7.1' ]]; then composer install -n ; fi
   - if [ "$dependencies" = "lowest" ]; then composer update --prefer-lowest --prefer-stable -n; fi;
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,10 @@ php:
   - 7.1
   - nightly
 
+cache:
+  directories:
+    - $HOME/.composer/cache
+
 matrix:
   fast_finish: true
   allow_failures:
--- a/composer.json
+++ b/composer.json
@@ -31,10 +31,10 @@
         "php-di/phpdoc-reader": "^2.0.1"
     },
     "require-dev": {
-        "phpunit/phpunit": "~5.0",
-        "mnapoli/phpunit-easymock": "~0.2.0",
-        "doctrine/annotations": "~1.2",
-        "ocramius/proxy-manager": "~2.0.2"
+        "phpunit/phpunit": "^5.7.0",
+        "mnapoli/phpunit-easymock": "^0.2.3",
+        "doctrine/annotations": "^1.2.0",
+        "ocramius/proxy-manager": "^2.0.2"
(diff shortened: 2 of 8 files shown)


mnapoli:
That is not really reassuring though: https://twitter.com/Ocramius/status/867477833123328000

> Yes, but then I would never use APC/APCu for anything under high write load for the same reason

https://twitter.com/Ocramius/status/867477517623599105

> Usually just the write being fast. APCu often gets cache corruption when under heavy write load, so we're under the same league

mnapoli:
Made obsolete by #494

Issue #152, opened 2014-02-05
Configuration dumper/debugger


mnapoli:
A nice idea [from reddit](http://www.reddit.com/r/PHP/comments/1x0r2f/phpdi_40_released/cf7mz57):
- dumping the definition of an entry to help debug it (CLI tool probably)
- dumping all the definitions in the container
- logging debug info (how the container resolves an entry)

holtkamp:
Just encountered this issue, we use separate definition files and then one can use (quite raw, but you get the idea):

```php
 $debugDicConfiguration = false;
    if ($debugDicConfiguration === true) {
        $registeredResources = array();
        foreach ($definitionsFiles as $definitionsFile) {
            $definitions = require $definitionsFile;
            foreach ($definitions as $registeredResource => $value) {
                if (in_array($registeredResource, $registeredResources)) {
                    echo "Error, duplicate Resource name $registeredResource";
                }
                $registeredResources[] = $registeredResource;
                $resource = $dicContainer->get($registeredResource);
                printf('<pre>Loading %s resulted in class: %s, last errors: %s</pre>', $registeredResource, is_object($resource) ? get_class($resource) : gettype($resource), print_r(error_get_last(), true));
            }
        }
        exit;
    }
```

juliangut:
@mnapoli how would you integrate _definition dumpers_ into a CLI tool such as a Symfony/Console? 

Just a single command to dump container configuration like so:
```
php bin/console container:debug
```

mnapoli:
@juliangut at the moment if you have a definition object, you can simply cast it to string and that will give you debug information. So you could write a CLI command that does that.

However the problem is getting the definition. There is no method on the container. And I'm not sure I want to add one :/ (because definitions are pretty "internal")

juliangut:
@mnapoli have you changed your mind about exposing definitions in order to create something like I commented? maybe not expose "exactly" internal definitions but some sort of simple definition object

mnapoli:
@juliangut yes I think that's a good idea. Since definitions are internal maybe we could have a `dumpDefinition($entryName) : string` method on the container. I could return the dumped definition as a string?

juliangut:
With your proposed `dumpDefinition($entryName)` I'd need to know all $entryName I want to dump, but what I want is to show whole Container definitions. I guess I could have an array of definitions prior to give it to ContainerBuilder so I can get the keys and then call dumpDefinition for each

[...]

PR #505, opened 2017-06-09
container entries debug

juliangut:
covers container entries debug dumper as per [comment on #152](https://github.com/PHP-DI/PHP-DI/issues/152#issuecomment-306288305)

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace DI;
 
 use DI\Definition\Definition;
+use DI\Definition\Exception\InvalidDefinition;
 use DI\Definition\FactoryDefinition;
 use DI\Definition\Helper\DefinitionHelper;
 use DI\Definition\InstanceDefinition;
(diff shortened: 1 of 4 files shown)


tests/UnitTest/ContainerGetTest.php
```diff
@@ -63,4 +64,78 @@ public function testPassByReferenceParameter()
...
+        $this->assertRegExp('/^Object \(\n {4}class = DI\\\Container\n/', $container->debugEntry('autowire'));
+        $this->assertEquals('Factory', $container->debugEntry('factory'));
+        $this->assertEquals('Decorate(decorator)', $container->debugEntry('decorator'));
+        $this->assertEquals('get(value)', $container->debugEntry('alias'));
+        $this->assertRegExp('/^Environment variable \(\n {4}variable = foo\n/', $container->debugEntry('environment'));
+        $this->assertEquals("[\n    0 => 'foo',\n]", $container->debugEntry('array'));
+        $this->assertEquals('foo', $container->debugEntry('string'));
+        $this->assertEquals('string', $container->debugEntry('entry'));
+        $this->assertEquals('stdClass', $container->debugEntry('object'));
+    }
```
mnapoli: This test is `ContainerGetTest.php` and it's supposed to test the `get()` method.

I think it would make more sense to move those 2 new tests into a new test class in `IntegrationTest` (instead of UnitTest) that would test the debug features, i.e. `DebugTest` ?

src/Container.php
```diff
@@ -267,6 +268,48 @@ public function set(string $name, $value)
     }
 
     /**
+     * Get defined container entries.
+     *
+     * @return array
+     */
+    public function getKnownEntryNames()
```
mnapoli: You can add the `array` return type here and remove the same line that is in the phpdoc (PHP 7 FTW ^^)

src/Container.php
```diff
@@ -267,6 +268,48 @@ public function set(string $name, $value)
...
+     * Get entry debug information.
+     *
+     * @param string $name Entry name
+     *
+     * @throws InvalidArgumentException
+     * @throws InvalidDefinition
+     *
+     * @return string
+     */
+    public function debugEntry($name)
```
mnapoli: Could you add the strict types here too? (parameter + return type)

src/Container.php
```diff
@@ -267,6 +268,48 @@ public function set(string $name, $value)
...
+            return (string) $definition;
+        }
+
+        if (array_key_exists($name, $this->resolvedEntries)) {
+            $entry = $this->resolvedEntries[$name];
+
+            return is_object($entry) ? get_class($entry) : gettype($entry);
+        }
+
+        throw new InvalidArgumentException("No entry or class found for '$name'");
```
mnapoli: You could throw a `NotFound` exception? (it wouldn't be a huge difference but since the exception exists let's use it)

src/Container.php
juliangut: I think code base must be reviewed on missing type hinting hunt, for example on [Container::has](https://github.com/PHP-DI/PHP-DI/blob/master/src/Container.php#L184)

mnapoli:
Thanks!

PR #507, opened 2017-06-11
Compile factories, including closures

mnapoli:
Factory definitions are now dumped into the compiled container!

This follows #494 where all definitions but factories, decorators and wildcards were compiled. Decorators and wildcards are still not compiled.

Compiling factories means also compiling closures. That was actually much more doable than I thought, I've used [Roave/BetterReflection](https://github.com/Roave/BetterReflection) (thanks @asgrim and @ocramius!)

The main downside of the whole pull request is the number of dependencies that BetterReflection brings:

  - Installing symfony/polyfill-util (v1.4.0) Loading from cache
  - Installing symfony/polyfill-php56 (v1.4.0) Loading from cache
  - Installing nikic/php-parser (v2.1.1) Loading from cache
  - Installing jeremeamia/superclosure (2.3.0) Loading from cache
  - Installing zendframework/zend-eventmanager (3.1.0) Loading from cache
  - Installing zendframework/zend-code (3.1.0) Loading from cache
  - Installing phpdocumentor/reflection-common (1.0) Loading from cache
  - Installing phpdocumentor/type-resolver (0.2.1) Loading from cache
  - Installing phpdocumentor/reflection-docblock (2.0.5) Loading from cache
  - Installing roave/better-reflection (1.2.0) Loading from cache

Maybe we could get on with superclosure only. Or else make compiling closures optional (only compile them if BetterReflection is installed…). Also I've seen that there is a v2.0 of BetterReflection in development [but it requires 7.1](https://github.com/Roave/BetterReflection#changes-in-br-20), I have no idea for how long the v1 will be supported (e.g. will it support PHP 7.2). PHP-DI could upgrade to 7.1 at the end of this year but that's not for sure yet.

Performance improvements: -20% on the factory.php benchmark, [profile](https://blackfire.io/profiles/compare/6692ca3e-521a-49df-96c6-1bf9833c8209/graph), which is very good for a micro-benchmark IMO. I'll test all that in larger scenarios for the complete 6.0 release but it's safe to say this PR is an improvement.

There are also a lot of micro/not-so-micro optimisations possible here in the future, especially because there are a lot of parameter-guessing done at runtime (you can inject dependencies into the factories and PHP-DI figures it out with type-hints). https://github.com/PHP-DI/Invoker could be compiled in the future (optionally), and simpler scenarios (classic closures) could be much more optimized than that. Work for later! :)

Because of closures, some scenarios are explicitly not supported:

- you should not use `$this` inside closures
- you should not import variables inside the closure using the `use` keyword, like in `function () use ($foo) { ...`
- you should not define multiple closures on the same line

But those scenarios don't make sense in container factories written in array config so that's good!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "psr/container": "^1.0",
         "php-di/invoker": "^2.0",
         "php-di/phpdoc-reader": "^2.0.1",
-        "roave/better-reflection": "^1.2",
+        "jeremeamia/superclosure": "^2.0",
         "nikic/php-parser": "^2.0|^3.0"
     },
     "require-dev": {
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,8 @@
         "php": ">=7.0.0",
         "psr/container": "^1.0",
         "php-di/invoker": "^2.0",
-        "php-di/phpdoc-reader": "^2.0.1"
+        "php-di/phpdoc-reader": "^2.0.1",
+        "roave/better-reflection": "^1.2.0"
     },
     "require-dev": {
         "phpunit/phpunit": "~5.7",
@@ -44,5 +45,10 @@
     "suggest": {
         "doctrine/annotations": "Install it if you want to use annotations (version ~1.2)",
         "ocramius/proxy-manager": "Install it if you want to use lazy injection (version ~2.0)"
+    },
+    "config": {
+        "platform": {
+            "php": "7.0"
+        }
     }
 }
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,8 @@
         "psr/container": "^1.0",
         "php-di/invoker": "^2.0",
         "php-di/phpdoc-reader": "^2.0.1",
-        "roave/better-reflection": "^1.2.0"
+        "roave/better-reflection": "^1.2",
+        "nikic/php-parser": "^2.0|^3.0"
     },
     "require-dev": {
         "phpunit/phpunit": "~5.7",
--- a/composer.json
+++ b/composer.json
@@ -45,10 +45,5 @@
     "suggest": {
         "doctrine/annotations": "Install it if you want to use annotations (version ~1.2)",
         "ocramius/proxy-manager": "Install it if you want to use lazy injection (version ~2.0)"
-    },
-    "config": {
-        "platform": {
-            "php": "7.0"
-        }
     }
 }
--- a/src/CompiledContainer.php
+++ b/src/CompiledContainer.php
@@ -4,7 +4,17 @@ declare(strict_types=1);
 
 namespace DI;
 
+use DI\Compiler\RequestedEntryHolder;
 use DI\Definition\Definition;
+use DI\Definition\Exception\InvalidDefinition;
+use DI\Invoker\FactoryParameterResolver;
+use Invoker\Exception\NotCallableException;
+use Invoker\Exception\NotEnoughParametersException;
+use Invoker\Invoker;
+use Invoker\InvokerInterface;
+use Invoker\ParameterResolver\AssociativeArrayResolver;
+use Invoker\ParameterResolver\NumericArrayResolver;
+use Invoker\ParameterResolver\ResolverChain;
 
 /**
  * Compiled version of the dependency injection container.
@@ -13,6 +23,11 @@ use DI\Definition\Definition;
  */
 abstract class CompiledContainer extends Container
 {
+    /**
+     * @var InvokerInterface
+     */
+    private $factoryInvoker;
+
     /**
      * {@inheritdoc}
      */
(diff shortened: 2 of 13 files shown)


composer.json
```diff
@@ -44,5 +45,10 @@
     "suggest": {
         "doctrine/annotations": "Install it if you want to use annotations (version ~1.2)",
         "ocramius/proxy-manager": "Install it if you want to use lazy injection (version ~2.0)"
+    },
+    "config": {
+        "platform": {
+            "php": "7.0"
```
Ocramius: uhmmmmmmmm

src/CompiledContainer.php
```diff
@@ -74,4 +89,33 @@ protected function setDefinition(string $name, Definition $definition)
...
         throw new \LogicException('You cannot set a definition at runtime on a compiled container. You can either put your definitions in a file, disable compilation or ->set() a raw value directly (PHP object, string, int, ...) instead of a PHP-DI definition.');
     }
+
+    /**
+     * Invoke the given callable.
+     */
+    protected function resolveFactory($callable, $entryName, array $extraParameters = [])
+    {
+        // Initialize the factory resolver
+        if (! $this->factoryInvoker) {
```
Ocramius: Make this a private getter

src/Compiler.php
```diff
@@ -139,6 +150,31 @@ private function compileDefinition(string $entryName, Definition $definition) :
                 $code = $compiler->compile($definition);
                 $code .= "\n        return \$object;";
                 break;
+            case $definition instanceof FactoryDefinition:
+                $value = $definition->getCallable();
+
+                // Custom error message to help debugging
+                $isInvokableClassName = is_string($value) && class_exists($value) && method_exists($value, '__invoke');
```
Ocramius: Method should not be static

Ocramius: Variable name does not reflect its meaning: this is not a name, it's a `bool`

src/Compiler.php
```diff
@@ -139,6 +150,31 @@ private function compileDefinition(string $entryName, Definition $definition) :
...
+                if ($isInvokableClassName && !$this->autowiringEnabled) {
+                    throw new InvalidDefinition(sprintf(
```
Ocramius: You should really start using named constructors for exceptions :-P

src/Compiler.php
```diff
@@ -182,6 +218,33 @@ public function compileValue($value) : string
             return "[\n$value        ]";
         }
 
+        if ($value instanceof \Closure) {
+            try {
+                $reflection = ReflectionFunction::createFromClosure($value);
+            } catch (TwoClosuresOneLine $e) {
+                throw new InvalidDefinition('Cannot compile closures when two closures are defined on the same line');
```
Ocramius: Do you really want to catch this? Also, make sure to re-throw the previous exception if you do.

src/Compiler.php
```diff
@@ -182,6 +218,33 @@ public function compileValue($value) : string
             return "[\n$value        ]";
         }
 
+        if ($value instanceof \Closure) {
```
Ocramius: Move to private method

src/Compiler.php
```diff
@@ -182,6 +218,33 @@ public function compileValue($value) : string
...
+            } catch (TwoClosuresOneLine $e) {
+                throw new InvalidDefinition('Cannot compile closures when two closures are defined on the same line');
+            }
+
+            /** @var Closure $ast */
+            $ast = $reflection->getAst();
+
+            // Force all closures to be static (add the `static` keyword), i.e. they can't use
+            // $this, which makes sense since their code is copied into another class.
+            $ast->static = true;
```
Ocramius: You are using the API from PHP-Parser here, so you should add that to your `composer.json` too

src/ContainerBuilder.php
```diff
@@ -144,7 +144,8 @@ public function build()
         $containerClass = $this->containerClass;
 
         if ($this->compileToFile) {
-            $containerClass = (new Compiler)->compile($source, $this->compileToFile);
+            $autowiringEnabled = $this->useAutowiring || $this->useAnnotations;
```
Ocramius: Variable is redundant

Ocramius:
> I have no idea for how long the v1 will be supported (e.g. will it support PHP 7.2). PHP-DI could upgrade to 7.1 at the end of this year but that's not for sure yet.

You'll get security support, no worries :-)

composer.json
mnapoli: What was I trying to do 🤔

src/CompiledContainer.php
mnapoli: I wanted to avoid an unnecessary method call here, see the profile of the difference: https://blackfire.io/profiles/compare/5a49dac2-a8f2-454d-9074-598540040477/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=focused&settings%5BtabPane%5D=nodes&selected=&callname=DI%5CCompiledContainer%3A%3AgetFactoryInvoker

Not much but if that can be avoided.

src/Compiler.php
mnapoli: I want to hide the implementation details of the closure compiler, but good catch on forwarding the previous exception.

src/Compiler.php
mnapoli: I'm not sure I should check for the staticness, [PHP already throws a warning at compile time](https://3v4l.org/nitUm) and [it still works](https://3v4l.org/Pb4bL) so I don't think PHP-DI should choose to make this stricter.

src/Compiler.php
mnapoli: I do for common ones but this error message is very specific to that line so moving it elsewhere would just spread logic around (cohesion ftw)

mnapoli:
Less Composer dependencies are now installed:

>  - Installing symfony/polyfill-util (v1.4.0): Loading from cache
>  - Installing symfony/polyfill-php56 (v1.4.0): Loading from cache
>  - Installing nikic/php-parser (v3.0.6): Loading from cache
>  - Installing jeremeamia/superclosure (2.3.0): Loading from cache

Ocramius:
@mnapoli would it be a good idea to have the factory compiling split out and handled as a dev dependency by consumer projects? Just wondering/rambling, not necessarily feasible or relevant.

mnapoli:
That could be a good idea indeed (considering the whole compilation step to make sense only in dev env, not prod) but I'd like to keep the whole thing simple (simply calling a method on the `ContainerBuilder` to compile the container is nice). The 4 extra dependencies are OK I think.

PR #508, opened 2017-06-15
Format debug dump entries

juliangut:
While definition's debug dump is well formatted, container entries debug dump is quite not that useful by just showing the type

I've tried to match entries format to definition formats as close as possible. This PR still needs a review to see if more types needs to be added or better formatting can be achieved, but you can get the idea

### DIFF
--- a/tests/IntegrationTest/ContainerDebugTest.php
+++ b/tests/IntegrationTest/ContainerDebugTest.php
@@ -109,7 +109,7 @@ class ContainerDebugTest extends BaseContainerTest
         $this->assertEquals('Value (100)', $container->debugEntry('entry_int'));
         $this->assertEquals('Value (false)', $container->debugEntry('entry_bool'));
         $this->assertEquals("Value ('string')", $container->debugEntry('entry_str'));
-        $this->assertEquals("Value (NULL)", $container->debugEntry('entry_null'));
+        $this->assertEquals('Value (NULL)', $container->debugEntry('entry_null'));
         $this->assertEquals('Value (Resource)', $container->debugEntry('entry_resource'));
         $this->assertEquals('Factory', $container->debugEntry('entry_callback'));
     }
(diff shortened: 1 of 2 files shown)


tests/IntegrationTest/ContainerDebugTest.php
```diff
@@ -73,17 +88,29 @@ public function testEntriesDefinitions()
...
-        $this->assertEquals('stdClass', $container->debugEntry('object'));
+        // Container entries
+        $this->assertEquals("Object (\n    class = stdClass\n)", $container->debugEntry('entry_object'));
+        $this->assertEquals("[\n    0 => 'foo',\n    1 => 'bar',\n]", $container->debugEntry('array'));
+        $this->assertEquals('Value (100)', $container->debugEntry('entry_int'));
+        $this->assertEquals('Value (false)', $container->debugEntry('entry_bool'));
+        $this->assertEquals("Value ('string')", $container->debugEntry('entry_str'));
+        $this->assertEquals("Value (NULL)", $container->debugEntry('entry_null'));
+        $this->assertEquals('Value (Resource)', $container->debugEntry('entry_resource'));
+        $this->assertEquals('Factory', $container->debugEntry('entry_callback'));
```
juliangut: I find this result a bit weird, I was expecting a 'class Closure' or 'callable'

mnapoli: That's "normal" in the sense that calling `$container->set()` with a closure turns it into a factory definition. If you get the entry you don't get the closure, you get what the closure returns. You have to use `DI\value()` with `$container->set()` if you want to preserve the value.

juliangut: I see, I've never set a closure on container (by set method) but defined factories as definitions

What would happen if I use DI functions with set method? i.e. `$container->set(\DI\get('key'))`

mnapoli: In that case it sets the definition in a definition source, so dumping/debugging will work just like if you put the same definition in an array. So you shouldn't worry about that case in your PR I think.

juliangut:
After reviewing I can see no other type to be added so I think it's good to go

mnapoli:
Awesome thanks!

PR #509, opened 2017-06-18
Rename the `compile()` method to `enableCompilation()`

### LINKED ISSUE #506: Rename `ContainerBuilder::compile()`
From https://github.com/PHP-DI/PHP-DI/pull/494#issuecomment-306310081

Suggestions:

- `setCompilationPath($filename)`: I'd like to avoid using setters if possible
- `enableCompilation($filename)`

> > Please note that the file name will also be the name of the generated PHP class. Because of that the filename you specify must also be a valid class name
>
> Wouldn't it be better just to use a single file name (CompiledContainer.php is just perfect) and let the user set the path only? (this plays better with my previous suggestion of setCompilationPath)

@juliangut I'm not sure I remember the whole reasons at the moment but I needed being able to set the file name for tests or scenarios where I wanted multiple containers. In any case I think it's also a good thing to let users choose the path and the file name, it allows to avoid creating a directory just for one file, and it allows more control over that file. Users will need to manage that file specifically for their deployment, some others may want to commit it in the repository. If we ever changed the file name it could have bad consequences, I think it's simpler and saner to let users choose the whole file name instead of forcing their hand.

mnapoli:
Fixes #506. Rename `ContainerBuilder::compile()` to `enableCompilation()`.

Parameters have also changed so that it's a bit simpler now.

**Before:**

```php
$containerBuilder->compile(__DIR__ . '/var/cache/CompiledContainer.php');
```

**After:**

```php
$containerBuilder->enableCompilation(__DIR__ . '/var/cache');

// You can optionally customize the generated class name:
$containerBuilder->enableCompilation(__DIR__ . '/var/cache', 'CompiledContainer');
```

TODO:

- [x] test that it does the job in the container benchmark

### DIFF
--- a/doc/performances.md
+++ b/doc/performances.md
@@ -31,7 +31,7 @@ $containerBuilder->enableCompilation(__DIR__ . '/var/cache');
 $container = $containerBuilder->build();
 ```
 
-The `enableCompilation()` method takes the name of the directory in which to store the compiled container.
+The `enableCompilation()` method takes the path of the directory in which to store the compiled container.
 
 ### Deployment in production
 
--- a/doc/performances.md
+++ b/doc/performances.md
@@ -20,29 +20,24 @@ In order to avoid those two tasks, the container can be compiled into PHP code o
 
 ### Setup
 
-Compiling the container is as easy as calling the `compile()` method on the container builder:
+Compiling the container is as easy as calling the `enableCompilation()` method on the container builder:
 
 ```php
 $containerBuilder = new \DI\ContainerBuilder();
-$containerBuilder->compile(__DIR__ . '/var/cache/CompiledContainer.php');
+$containerBuilder->enableCompilation(__DIR__ . '/var/cache');
 
 // […]
 
 $container = $containerBuilder->build();
 ```
 
-The `compile()` method takes a single argument: the name of a file in which to store the container.
-
-Please note that the file name will also be the name of the generated PHP class. Because of that the filename you specify must also be a valid class name. For example:
-
-- `var/cache/CompiledContainer.php`: valid
-- `var/cache/compiled-container.php`: invalid since `compiled-container` is not a valid PHP class name
+The `enableCompilation()` method takes the name of the directory in which to store the compiled container.
 
 ### Deployment in production
 
 When a container is configured to be compiled, **it will be compiled once and never be regenerated again**. That allows for maximum performances in production.
 
-When you deploy new versions of your code to production **you must delete the generated file** to ensure that the container is re-compiled.
+When you deploy new versions of your code to production **you must delete the generated file** (or the directory that contains it) to ensure that the container is re-compiled.
 
 If your production handles a lot of traffic you may also want to generate the compiled container *before* the new version of your code goes live. That phase is known as the "warmup" phase. To do this, simply create the container (call `$containerBuilder->build()`) during your deployment step and the compiled container will be created.
 
@@ -53,7 +48,7 @@ If your production handles a lot of traffic you may also want to generate the co
 ```php
 $containerBuilder = new \DI\ContainerBuilder();
 if (/* is production */) {
-    $containerBuilder->compile(__DIR__ . '/var/cache/CompiledContainer.php');
+    $containerBuilder->enableCompilation(__DIR__ . '/var/cache');
 }
 ```
 
--- a/src/Compiler.php
+++ b/src/Compiler.php
@@ -50,7 +50,7 @@ class Compiler
      */
     public function compile(DefinitionSource $definitionSource, string $directory, string $className) : string
     {
-        $fileName = $directory . '/' . $className . '.php';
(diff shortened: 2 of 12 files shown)


src/Compiler.php
```diff
@@ -46,21 +46,15 @@ class Compiler
...
+    public function compile(DefinitionSource $definitionSource, string $directory, string $className) : string
     {
-        $this->containerClass = basename($fileName, '.php');
-
-        // Validate that it's a valid class name
-        $validClassName = preg_match('/^[a-zA-Z_][a-zA-Z0-9_]*$/', $this->containerClass);
-        if (!$validClassName) {
-            throw new InvalidArgumentException("The file in which to compile the container must have a name that is a valid class name: {$this->containerClass} is not a valid PHP class name");
-        }
+        $fileName = $directory . '/' . $className . '.php';
```
sagikazarmark: Taking care of trailing slash might be a good idea here: `rtrim($directory, '/')`

src/ContainerBuilder.php
```diff
@@ -158,25 +159,27 @@ public function build()
...
+     * @param string $className Name of the class. Customize only if necessary.
      */
-    public function compile(string $fileName) : ContainerBuilder
+    public function enableCompilation(string $directory, string $className = 'CompiledContainer') : ContainerBuilder
     {
         $this->ensureNotLocked();
 
-        $this->compileToFile = $fileName;
+        $this->compileToDirectory = $directory;
+        $this->containerClass = $className;
```
sagikazarmark: You mentioned class name validation earlier, but I can't find any kind of validation.

src/ContainerBuilder.php
```diff
@@ -299,7 +302,7 @@ public function addDefinitions($definitions) : ContainerBuilder
      */
     public function isCompiled() : bool
     {
-        return (bool) $this->compileToFile;
+        return (bool) $this->compileToDirectory;
```
sagikazarmark: I am a little confused by this method. The docblock says "Will it be...", the method name says "Is it already...." and the body does a weird casting of a directory path variable. While I can guess what this method does with a high chance of hit, some clarification would not harm IMO.

tests/IntegrationTest/BaseContainerTest.php
```diff
@@ -12,41 +12,44 @@
...
 abstract class BaseContainerTest extends TestCase
 {
-    const COMPILED_CONTAINER_DIRECTORY = __DIR__ . '/tmp';
+    const COMPILATION_DIR = __DIR__ . '/tmp';
 
     public static function setUpBeforeClass()
     {
         // Clear all files
-        array_map('unlink', glob(self::COMPILED_CONTAINER_DIRECTORY . '/*'));
+        array_map('unlink', glob(self::COMPILATION_DIR . '/*'));
```
sagikazarmark: What do you think about adding vfs for these tests?

src/ContainerBuilder.php
mnapoli: Yes I deleted it, thinking that it's a pretty obvious error and that validating the class name with a regex on every HTTP request is a bit too much. Maybe I could put the validation at compile time only though, I'll have a look.

tests/IntegrationTest/BaseContainerTest.php
mnapoli: Why not but it already works fine, and it's just 1 line of code to clear the directory so I'm not sure it's worth it?

doc/performances.md
```diff
@@ -20,29 +20,24 @@ In order to avoid those two tasks, the container can be compiled into PHP code o
...
-- `var/cache/CompiledContainer.php`: valid
-- `var/cache/compiled-container.php`: invalid since `compiled-container` is not a valid PHP class name
+The `enableCompilation()` method takes the name of the directory in which to store the compiled container.
 
 ### Deployment in production
 
 When a container is configured to be compiled, **it will be compiled once and never be regenerated again**. That allows for maximum performances in production.
 
-When you deploy new versions of your code to production **you must delete the generated file** to ensure that the container is re-compiled.
+When you deploy new versions of your code to production **you must delete the generated file** (or the directory that contains it) to ensure that the container is re-compiled.
```
mnapoli: It is mentioned in the paragraph below, do you think that does the job or it's worth expanding more?

mnapoli:
Thanks for the review I'll have a look!

tests/IntegrationTest/BaseContainerTest.php
sagikazarmark: Makes the tests more stable (less fragile, susceptible to environment changes) IMO, but not a killer requirement.

doc/performances.md
sagikazarmark: OMG, I missed it, that's perfect. :man_facepalming:

src/ContainerBuilder.php
sagikazarmark: Probably validate only when actually compiling the container? A compiled container created in any other way is already considered reckless activity.

PR #511, opened 2017-06-21
Update annotations.md

rquadling:
PHP 7.0 dropped the `opcache.load_comments` setting. Comments, if saved, are now always loaded. But, without them being saved, they cannot be loaded, so adjust the troubleshooting guide to reflect this requirement.

http://php.net/ChangeLog-7.php#7.0.0 (scroll down to Opcache).
http://php.net/manual/en/opcache.configuration.php#ini.opcache.save-comments


mnapoli:
Thank you for that!

PR #514, opened 2017-07-17
create proxies directory

juliangut:
In the same fashion compilation directory is created if doesn't exist

### DIFF
--- a/src/Proxy/ProxyFactory.php
+++ b/src/Proxy/ProxyFactory.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace DI\Proxy;
 
+use InvalidArgumentException;
 use ProxyManager\Configuration;
 use ProxyManager\Factory\LazyLoadingValueHolderFactory;
 use ProxyManager\FileLocator\FileLocator;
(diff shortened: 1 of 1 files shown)


mnapoli:
Thanks! Though:

- this code will be executed on every http request (the compilation directory is only created at compile time) so it might be a performance impact: is that code worth it? I'm not sure since that's a pretty obvious error and easy to solve (especially since it's a feature that only most advanced users will use)
- it's specific to the implementation but `mkdir` is executed every time: there is no test if the directory exists beforehand?

juliangut:
If createProxiesDirectory method verifies directory existence prior to creating it then there would not be any performance issue, both problems solved at once

mnapoli:
Moving the creation in the `ContainerBuilder` will still execute it every time.

> If createProxiesDirectory method verifies directory existence prior to creating it then there would not be any performance issue

Testing if the directory exists still does a file access.

To be honest when you weight the pros and the cons I think it's OK to let the error happen, at worst it will happen once when the user will configure their deployment script.

juliangut:
I'm currently checking directory before setting it so it's OK for me

PR #521, opened 2017-08-05
Fix #520: Error if an optional parameter is before required parameters

### LINKED ISSUE #520: DI doesn't resolve, if an optional parameter comes before other definitions
For example there's some legacy code in a project I'm working on:

    Object (
        class = LanguageWebService
        scope = singleton
        lazy = false
        __construct(
            $language_directory = #UNDEFINED#
            $request = get(Symfony\Component\HttpFoundation\Request)
            $permissions = get(Permissions)
        )
    )

Even if the $language_directory is set to null by default (optional) php-di doesn't properly pass a optional null value when trying to resolve the dependency. If $language_directory is the last parameter it works as normal.

mnapoli:
Fix #520

### DIFF
--- a/.styleci.yml
+++ b/.styleci.yml
@@ -14,3 +14,4 @@ disabled:
     - phpdoc_align
     - phpdoc_separation
     - align_double_arrow
+    - no_unreachable_default_argument_value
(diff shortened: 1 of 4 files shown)


mnapoli:
Ha :) StyleCI fixed the broken test case because of https://styleci.readme.io/v1.0/docs/fixers#section-no_unreachable_default_argument_value

The last commit should do.

Issue #531, opened 2017-10-30
Suggestion to increase performance of $container->get(), $container->has() and $container->make()


holtkamp:
When in the process of increasing performance of our application, I noticed two things:

- fetched Definitions are not (locally) cached, for instance in a local array property `$acquiredDefinitions[$name] => $definition`
- array_key_exists() is used, while isset() [tends to perform better](https://ilia.ws/archives/247-Performance-Analysis-of-isset-vs-array_key_exists.html), although:
  - it is considered a micro-optimization...
  - when the value equals `NULL` it will be considered 'not set'

What do you think? Would a PR on this make sense? It can be as easy as:

```php
private function getDefinition($name)
{
    if(!array_key_exists($name, $this->acquiredDefinitions)){
        $this->acquiredDefinitions[$name] = $this->definitionSource->getDefinition($name);
    }

    return  $this->acquiredDefinitions[$name];
}
```
Maybe `$lookedUpDefinitions` or `$fetchedDefinitions` is a better name for the property...

Attached a part of a BlackFire.io profiling session of the current situation.

![screen shot 2017-10-30 at 13 44 36](https://user-images.githubusercontent.com/776405/32173660-02d2bf26-bd80-11e7-913a-5b08e7f4ec04.png)

 Adding this local cache reduces the amount of lookups with around 390 calls.

mnapoli:
Hi, thanks for all those details and the pull request. Have you tried an alpha release of the future v6? It is meant to be much faster (the container is compiled), so I'm not sure this optimisation is really necessary.

holtkamp:
>Have you tried an alpha release of the future v6? 

Yeah, tried it a while back, but back then the migration effort was too high + v5 is quite stable and "works for me" 😄

mnapoli:
Thanks that's valuable feedback, do you remember what was causing the effort? I can probably add a few helpers or maybe documentation in the v6 release to help others that are in your situation.

Anyway regarding v5 I guess we could merge your pull request in a new v5 release but not port it to v6 as to avoid complicating the code for no gain?

holtkamp:
> Thanks that's valuable feedback, do you remember what was causing the effort?

Mmm, don't remember, maybe I can revive this `PHP-DI-6` feature-branch of our project somewhere this week.

> Anyway regarding v5 I guess we could merge your pull request in a new v5 release but not port it to v6 as to avoid complicating the code for no gain?

[...]

PR #533, opened 2017-10-30
Container::getDefinition() which uses a local cache

holtkamp:
As suggested in https://github.com/PHP-DI/PHP-DI/issues/531

### DIFF
--- a/benchmarks/DI/ContainerBench.php
+++ b/benchmarks/DI/ContainerBench.php
@@ -1,0 +1,70 @@
+<?php
+
+namespace DI;
+
+use Psr\Container\ContainerInterface;
+use stdClass;
+
+/**
+ * @BeforeMethods({"initContainer"})
+ */
+class ContainerBench
+{
+    /**
+     * @var ContainerInterface | FactoryInterface
+     */
+    private $container;
+
+    public function initContainer()
+    {
+        $builder = new ContainerBuilder();
+
+        $builder->addDefinitions([
+            stdClass::class => function (ContainerInterface $container) {
+                return new stdClass();
+            },
+        ]);
+
+        $extraDefinitionsToSimulateRealisticContainer = array();
+        for ($i = 0; $i < 1000; $i++) {
+            $extraDefinitionsToSimulateRealisticContainer[stdClass::class . $i] = function (ContainerInterface $container) {
+                return new stdClass();
+            };
+        }
+        $builder->addDefinitions($extraDefinitionsToSimulateRealisticContainer);
+        $this->container = $builder->build();
+    }
+
+    /**
+     * @Revs(10000)
+     */
+    public function benchGetDummyClass()
+    {
+        $value = $this->container->get(stdClass::class);
+    }
+
+    /**
+     * @Revs(10000)
+     */
+    public function benchCheckAndGetDummyClass()
+    {
+        $value = $this->container->has(stdClass::class) ? $this->container->get(stdClass::class) : null;
+    }
+
+    /**
+     * @Revs(10000)
+     */
+    public function benchMakeDummyClass()
+    {
+        $value = $this->container->make(stdClass::class);
+    }
+
+
+    /**
+     * @Revs(10000)
+     */
+    public function benchCheckAndMakeDummyClass()
+    {
+        $value = $this->container->has(stdClass::class) ? $this->container->make(stdClass::class) : null;
+    }
+}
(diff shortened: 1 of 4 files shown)


avant1:
As I can see this breaks ability to update definition using `set()` after it was requested the first time.
Example:

```php
$myService = $container->get(MyService::class);
$container->set(MyService::class, new MyServiceDecorator($myService));

//$myUpdatedService === $myService
//set() has no effect
$myUpdatedService = $container->get(MyService::class);
```

`$fetchedDefinitions` could be cleared during `set()` call.

holtkamp:
>As I can see this breaks ability to update definition using set()

Aah, indeed, never knew that was possible. @mnapoli should I update the PR, or would you prefer to not merge it at all since focus is on v6?

src/DI/Container.php
```diff
@@ -49,6 +49,13 @@ class Container implements ContainerInterface, InteropContainerInterface, Factor
     private $definitionResolver;
 
     /**
+     * Map of definitions that are already fetched / looked up.
+     *
+     * @var array
+     */
+    private $fetchedDefinitions = [];
```
sagikazarmark: How about calling it `resolvedDefinitions` or `definitionCache`.

Also I would mention the purpose here: by caching the resolved container entries we gain performance.

src/DI/Container.php
```diff
@@ -306,7 +322,10 @@ private function setDefinition($name, Definition $definition)
             throw new \LogicException('The container has not been initialized correctly');
         }
 
-        // Clear existing entry if it exists
+        // Clear existing entries if it exists
+        if (array_key_exists($name, $this->fetchedDefinitions)) {
```
sagikazarmark: To be honest I think this is not enough to mitigate the stale cache problem.

What if you set a transitive dependency? (A depends on B, resolve A, set another B, resolve A -> I expect the new B, but I still get the old one?).

holtkamp: so after a 'set' all local 'caches' should be totally cleared?

src/DI/Container.php
holtkamp: yeah, I considered `resolvedDefinitions` as well, but the rest of the code kind of uses 'resolving' for 'resolving a value from a definition'. 

`definitionCache` could be candidate... @mnapoli ?

mnapoli:
Current status: discussion in #531 about this pull request ;)

src/DI/Container.php
mnapoli: :+1: to clear all caches, using `set()` is not the best option for performances anyway, it is more important that it works correctly.

mnapoli: Could you also try to add a test for that case? I'm not sure if it will be easy though

src/DI/Container.php
mnapoli: 👎  for `resolved`, it means something different indeed.

👍  for `definitionCache`

benchmarks/DI/ContainerBench.php
```diff
@@ -0,0 +1,70 @@
...
+
+
+    /**
+     * @Revs(10000)
+     */
+    public function benchCheckAndMakeDummyClass()
+    {
+        $value = $this->container->has(stdClass::class) ? $this->container->make(stdClass::class) : null;
+    }
+}
```
mnapoli: For performance tests there is already the folder `tests/PerformanceTest`. I run them with Blackfire so then I can compare the time taken and see if there are improvements. If it's not necessary to introduce another tool I'd rather keep it simple? (but feel free to add a new performance test file though)

holtkamp: Well, PHPBench is more fore benchmarking, while Blackfire is for profiling. It is quite a light-weight thingie that is only required in "development mode", I would advise to adopt it for these exact purposes.

sagikazarmark: There is already quite a good set of benchmarks with different containers here: https://github.com/kocsismate/php-di-container-benchmarks

I like PHPBench, but not for these kind of things where autoload and opcache also matter.

Also I don't really see the value in benchmarks if there is nothing to compare with.

My 2 cents.

holtkamp: Mmm, ok, I thought it was useful. Maybe the maintainers can cherry-pick the commits they deem useful and discard the rest 😉

mnapoli:
Thanks!

Issue #535, opened 2017-10-31
Allow extending \DI\CompiledContainer


juliangut:
I'm currently stuck with a problem with the compiled container:

I've made an extension over \DI\Container to access array entries by dot notation. Thanks to the ability to [set container class](https://github.com/PHP-DI/PHP-DI/blob/master/src/ContainerBuilder.php#L103) on ContainerBuilder it is a breeze to use my modified container

The problem comes when I compile the container (definitions), even though [compiled container template](https://github.com/PHP-DI/PHP-DI/blob/master/src/Compiler/Template.php#L4) allows to set file/class name it still hardcodes \DI\ContainerInterface as the parent class to it

It would be great not to [override container class](https://github.com/PHP-DI/PHP-DI/blob/master/src/ContainerBuilder.php#L187) but to set both compiled class name and compiled class parent name

Unless I'm missing something this should be a simple/quick PR

Edit: typos

mnapoli:
Nice catch, and thanks for the PR!

PR #536, opened 2017-10-31
configure compiled container parent class

juliangut:
Covers #535 why wait? 😄


mnapoli:
Awesome thanks, I wanted to push a test in your branch but wasn't allowed to push to your repository so I've opened #538 (it preserves your commit). When I merge it it should mark this one as merged.

PR #538, opened 2017-11-02
Second take for #536 "configure compiled container parent class"

### LINKED ISSUE #535: Allow extending \DI\CompiledContainer

mnapoli:
Fix #535

ping @juliangut 

This PR includes #536 + tests.


src/Compiler.php
```diff
@@ -56,8 +61,13 @@ class Compiler
      *
      * @return string The compiled container file name.
      */
-    public function compile(DefinitionSource $definitionSource, string $directory, string $className, bool $autowiringEnabled) : string
-    {
+    public function compile(
+        DefinitionSource $definitionSource,
+        string $directory,
+        string $className,
+        string $parentClassName,
```
bradynpoulsen: This is a breaking change in the API. Is method intend to be internal or public?

src/Compiler/Template.php
```diff
@@ -1,7 +1,7 @@
 /**
  * This class has been auto-generated by PHP-DI.
  */
-class <?=$this->containerClass; ?> extends \DI\CompiledContainer
+class <?=$this->containerClass; ?> extends <?=$this->containerParentClass; ?>
```
bradynpoulsen: Are compiled containers supposed to honor the Container Interface? If so, it seems like allowing a custom parent should either validate such either in the compile/enableCompilation methods or here via `implements`

src/Compiler.php
mnapoli: No problem with BC here:

- the class was added in 6.0 and we are still in the alpha release cycle
- the class is not tagged with `@api` so it's an internal class

src/Compiler/Template.php
mnapoli: Mmh this is a good question… Honestly I don't really mind what happens if someone decides to do something weird here?

juliangut: Interesting point here, I already considered this and I'm checking if the class [extends from \DI\CompiledContainer](https://github.com/juliangut/slim-php-di/blob/2.x/src/Configuration.php#L318)

Anyway I'm with @mnapoli, if you use any other thing than a ContainerInterface ... it's just a bug, fix it

PR #540, opened 2017-11-04
Rework how nested definitions are handled (fix #501 & #487)

### LINKED ISSUE #487: Always interpret closures as factories
When a factory is at the root of the definition array, it is automatically interpreted as a factory (`DI\factory()` is not necessary).

However it's not the case if the closure is in an array or an object parameter.

We could simplify the behavior and always consider a closure as a factory. If a user wants to really inject a closure, `DI\value()` is here for that. The behavior will at least be more consistent and simple.

### LINKED ISSUE #501: Nested autowire does not work
Not sure if it is intended/limitation, nevertheless I couldn't find any mention in the documentation (could be I didn't read it carefully).

Reproduced here: https://github.com/reproduce/php-di/tree/nested_autowire

mnapoli:
Sorry for the big PR if you intend to review, the work behind that kinda grow organically from a small change :) Reworking those commits would be very complicated.

This PR solves the problem of **nested definitions**, i.e.:

- fixes #501: support nested autowire (and includes #503 which were the tests), some examples:
    ```php
        [
            Foo::class => create()
                ->constructor(autowire(Bar::class)),
            'array' => [
                autowire(...),
                autowire(...),
            ],
        ]
    ```

- fixes #487: support nested anonymous functions (aka factories), some examples:

    ```php
        [
            Foo::class => create()
                ->constructor(function () { return ...; }),
            'array' => [
                function () { return ...; },
                function () { return ...; },
            ],
            'env' => \DI\env('FOO', function () { return ...; }),
        ]
    ```
- removes dead code
- simplifies the code
- removes some useless classes (some `DefinitionHelper` classes) because we can use `Definition` classes directly in most cases, that will mean less code to execute at runtime and less code to maintain

The main change is this PR is very structural:

- before: definitions were returned by "sources" (definition files, autowiring, etc.), when they were resolved the "resolver" would handle nested definitions
- problem: the "resolver" is not able to handle all cases correctly, and its job is also now duplicated in the compiler (that creates the compiled and optimized container)
- solution: definition "sources" now process (called **normalize** in the code) definitions and their nested definitions so that they are all clean before going to the resolver or compiler

(to normalize definitions they are now like "tree nodes", i.e. the normalizer visits each definition and its subdefinitions to prepare it for the resolver)

That means a bit more job in definition sources, less in resolvers and compiler. The code is simpler and handles all edge cases the same way \o/

TODO:

- [x] performance tests
- [x] changelog
- [x] documentation


src/Compiler.php
```diff
@@ -218,10 +217,6 @@ private function compileDefinition(string $entryName, Definition $definition) :
 
     public function compileValue($value) : string
     {
-        if ($value instanceof DefinitionHelper) {
-            $value = $value->getDefinition('');
-        }
```
mnapoli: cf. the description of the PR but there are no `DefinitionHelper` anymore once a definition is returned by the `DefinitionSource` (the definition is 100% prepared by the source, meaning less work for the compiler and definition resolvers).

src/Definition/ArrayDefinition.php
```diff
@@ -18,19 +16,15 @@ class ArrayDefinition implements Definition
...
 
     /**
      * @var array
      */
     private $values;
 
-    /**
-     * @param string $name Entry name
-     */
-    public function __construct(string $name, array $values)
```
mnapoli: There will be the same pattern on several definition classes: by making the name optional that allows me to get rid of several `DefinitionHelper` classes. The name is simply injected later by the `DefinitionSource`.

It's a tiny bit less clean, but it removes a lot of code and classes, it simplifies everything and also it makes a little sense because nested definitions don't have names, so… the name is not really mandatory after all.

src/Definition/Definition.php
```diff
@@ -21,6 +21,16 @@
     public function getName() : string;
 
     /**
+     * Set the name of the entry in the container.
+     */
+    public function setName(string $name);
```
mnapoli: This is the new method that allows to get rid of several `DefinitionHelper` classes (see the comment above).

src/Definition/Definition.php
```diff
@@ -21,6 +21,16 @@
...
+
+    /**
+     * Apply a callable that replaces the definitions nested in this definition.
+     */
+    public function replaceNestedDefinitions(callable $replacer);
```
mnapoli: This is the method that allows the new `DefinitionNormalizer` to traverse all nested definitions to prepare them (e.g. turn closures into "factory" definitions, process `autowire()` definition helpers, etc.)

src/Definition/Dumper/ObjectDefinitionDumper.php
```diff
@@ -109,11 +109,13 @@ private function dumpMethodParameters(string $className, MethodInjection $method
             if (array_key_exists($index, $definitionParameters)) {
                 $value = $definitionParameters[$index];
 
-                if ($value instanceof EntryReference) {
-                    $args[] = sprintf('$%s = get(%s)', $parameter->getName(), $value->getName());
+                if ($value instanceof Definition) {
+                    $valueStr = (string) $value;
```
mnapoli: All these little changes also mean that nested definitions are better supported when debugging and in exception messages.

mnapoli:
Performance tests: I see no difference, both when the container is compiled or not. So no performance impact and a simpler codebase: ✅

sagikazarmark:
Awesome!

PR #551, opened 2017-11-11
Update to PHPUnit 6

gabriel-caruso:
As we are support `PHP 7`, and the `mnapoli/phpunit-easymock` updated, now we can update our suite of tests to [`PHPUnit 6`](https://github.com/sebastianbergmann/phpunit/blob/master/ChangeLog-6.0.md#600---2017-02-03). I used [this article](https://thephp.cc/news/2017/02/migrating-to-phpunit-6) from @sebastianbergmann to help me in this upgrade.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,8 +32,8 @@
         "nikic/php-parser": "^2.0|^3.0"
     },
     "require-dev": {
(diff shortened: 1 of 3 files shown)


mnapoli:
Thanks!

PR #555, opened 2017-11-24
Fix make() fails when using decorate

### LINKED ISSUE #554: make() fails when using decorate
Hi @mnapoli 

I am getting currently this error with PHP-DI 5.4.5:

```
Entry "Piwik\Plugins\CoreHome\FooBar" cannot be resolved: Parameter $customer of __construct() has no value defined or guessableFull definition:Object ( class = Piwik\Plugins\CoreHome\FooBar scope = singleton lazy = false __construct( $customer = #UNDEFINED# ))
```

when calling:
```php
            return $container->make('Piwik\Plugins\CoreHome\FooBar', array(
                'customer' => $customer
            ));
```

and having a DI config like:

```php
'Piwik\Plugins\CoreHome\FooBar' => DI\decorate(function ($customer, \Interop\Container\ContainerInterface $c) {

        $myTest = $c->get('test.vars.myTest');
        if (!empty($myTest)) {
            return new MyTest();
        }

        return $customer;
    }),
```

The class looks like this:
```
class FooBar {
    /**
     * @var array
     */
    private $customer;

    public function __construct($customer)
    {
        $this->customer = $customer;
    }
```

It works nicely as soon as I remove the `DI\decorate()` but not when I add it. It does not even go into the decorator when debugging it.

It fails in `ObjectCreator::createInstance()` with a `DefinitionException`.

It seems like the passed parameters get lost in `DecoratorResolver::resolve(Definition $definition, array $parameters = [])` where the $parameters are actually never used.

tsteur:
fixes #554

@mnapoli looks like this may fix my issue


mnapoli:
Thanks, do you need this resolved in 5.x or future 6.0? `master` is currently for the next major release, feel free to change the target branch for the PR (you will need to rebase against `5.4`).

tsteur:
Changed the base but that wasn't a good idea :) I will just issue a new PR :)

PR #556, opened 2017-11-24
Fix make() fails when using decorate

### LINKED ISSUE #554: make() fails when using decorate

tsteur:
fixes #554

@mnapoli looks like this might fix the issue

### DIFF
--- a/src/DI/Definition/Resolver/DecoratorResolver.php
+++ b/src/DI/Definition/Resolver/DecoratorResolver.php
@@ -71,7 +71,7 @@ class DecoratorResolver implements DefinitionResolver
             ));
         }
 
-        $decorated = $this->definitionResolver->resolve($decoratedDefinition);
+        $decorated = $this->definitionResolver->resolve($decoratedDefinition, $parameters);
 
         return call_user_func($callable, $decorated, $this->container);
     }
--- a/tests/UnitTest/ContainerMakeTest.php
+++ b/tests/UnitTest/ContainerMakeTest.php
@@ -4,6 +4,7 @@ namespace DI\Test\UnitTest;
 
 use DI\ContainerBuilder;
 use DI\Test\UnitTest\Fixtures\Class1CircularDependencies;
+use DI\Test\UnitTest\Fixtures\Foo;
 use DI\Test\UnitTest\Fixtures\InvalidScope;
 use DI\Test\UnitTest\Fixtures\PassByReferenceDependency;
 use DI\Test\UnitTest\Fixtures\Prototype;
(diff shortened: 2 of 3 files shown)


mnapoli:
Thanks, would you have time to add tests? (https://github.com/PHP-DI/PHP-DI/blob/5.4/tests/UnitTest/ContainerMakeTest.php might be a good place in this branch) Sorry I did not check that earlier.

tsteur:
I tried to write a test but failed.

In https://github.com/PHP-DI/PHP-DI/blob/5.4/tests/UnitTest/ContainerMakeTest.php I couldn't figure out on how to add a decorator to the already built dev container. Then wanted to add a test to https://github.com/PHP-DI/PHP-DI/blob/5.4/tests/UnitTest/Definition/Resolver/DecoratorResolverTest.php
like this:

```php


    /**
     * @test
     */
    public function should_resolve_parameters()
    {
        $previousDefinition = new ObjectDefinition(PassByReferenceDependency::class);

        $callable = function ($previous, ContainerInterface $container) {
            return $previous->foo .= 'baz';
        };
        $definition = new DecoratorDefinition(PassByReferenceDependency::class, $callable);
        $definition->setSubDefinition($previousDefinition);

        $stdClass = new stdClass();
        $object = $this->resolver->resolve($definition, array('object' => $stdClass));
        $this->assertEquals('barbaz', $object->foo);
    }
```

But there are too many classes involved and I can't make it work :) No clue on how those definitions should be composed etc.

mnapoli:
I tried to push a commit to this branch but I am not allowed, here is the test I wrote:

```php
    /**
     * @see https://github.com/PHP-DI/PHP-DI/issues/554
     */
    public function testMakeWithDecorator()
    {
        $builder = new ContainerBuilder;
        $builder->addDefinitions([
            Foo::class => \DI\decorate(function ($previous) {
                return $previous;
            }),
        ]);
        $result = $container->make(Foo::class, [
            'bar' => 'baz',
        ]);
        $this->assertEquals('baz', $result->bar);
    }
```

With this fixture class:

```php
namespace DI\Test\UnitTest\Fixtures;

class Foo
{
    public $bar;

    public function __construct($bar)
    {
        $this->bar = $bar;
    }
}
```

Let me know if that sounds like your problem (it does fail on 5.4).

tsteur:
Pushed the test 👍  Thanks a lot. Edits should be allowed btw.

mnapoli:
Thanks!

PR #557, opened 2017-11-25
isCompiled

EgorGruzdev:
Example:
```php
if(!$containerBuilder->isCompiled()){
      $containerBuilder->addDefinitions(__DIR__ . '/../config/plugins/laravel.database.php');
      $containerBuilder->addDefinitions(__DIR__ . '/../config/application.php');
      $containerBuilder->addDefinitions(__DIR__ . '/../config/database.php');
}
```

### DIFF
--- a/src/ContainerBuilder.php
+++ b/src/ContainerBuilder.php
@@ -318,7 +318,20 @@ class ContainerBuilder
      */
     public function isCompiled() : bool
     {
-        return (bool) $this->compileToDirectory;
+        if(null === $this->compileToDirectory){
+            return false;
+        }
+
+        if(class_exists($this->containerClass)){
+            return true;
+        }else{
+            $fileName = rtrim($this->compileToDirectory, '/') . '/' . $this->containerClass . '.php';
+            if (file_exists($fileName)) {
+                return true;
+            }
+        }
+
+        return false;
     }
 
     private function ensureNotLocked()
--- a/src/ContainerBuilder.php
+++ b/src/ContainerBuilder.php
@@ -325,7 +325,7 @@ class ContainerBuilder
         if (class_exists($this->containerClass)) {
             return true;
         } else {
-            $fileName = rtrim($this->compileToDirectory, DIRECTORY_SEPARATOR);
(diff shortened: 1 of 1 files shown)


u-mulder:
- Code does not follow PSR coding standards.
- Instead of slashes use `DIRECTORY_SEPARATOR` constant

mnapoli:
I do not understand what this pull request is meant to achieve. Could you please explain, and also add tests (on top of @u-mulder's remarks).

EgorGruzdev:
If the container is already compiled, then why do I connect the files again with "addDefinitions".

To do this, and made a modification, if the compiler is already compiled, then do not load information from the connected files.

In the old version of this method, the check was based on data that there is an indication that you need to compile, and in fact there was not a compilation.

mnapoli:
I'm sorry I still don't understand.

Maybe the confusion comes from the fact that this is the `ContainerBuilder` class, not `Container`: `isCompiled()` returns `true` if the container will be compiled. The ContainerBuilder is the class that let us build/configure the container, so the container does not exist yet.

EgorGruzdev:
When working with compilation enabled, there was a problem that even if the compiled container exists, using the addDefinitions method, the file is connected.

Example:
'''php
$containerBuilder = new ContainerBuilder;
$containerBuilder->enableCompilation(__DIR__);
//If the container is compiled (compiled), then it makes no sense to call addDefinitions
if ($containerBuilder->isCompiled()) {
  $containerBuilder->addDefinitions(__DIR__ . '/app.config.php');
  $containerBuilder->addDefinitions(__DIR__ . '/module1.config.php');
  $containerBuilder->addDefinitions(__DIR__ . '/module2.config.php');
  $containerBuilder->addDefinitions(__DIR__ . '/module3.config.php');
  $containerBuilder->addDefinitions(__DIR__ . '/module4.config.php');
  $containerBuilder->addDefinitions(__DIR__ . '/module5.config.php');
  //....
}
$this->configureContainer($containerBuilder);
'''
My implementation of the method  "isCompiled" makes it possible to avoid repeated loading of files if the container is already compiled.

P.S. Sorry for my english

PR #559, opened 2017-11-30
Use assertNull

carusogabriel:
I've refactored some tests using the [`assertNull`](https://phpunit.de/manual/current/pt_br/appendixes.assertions.html#appendixes.assertions.assertNull) method.

### DIFF
--- a/tests/UnitTest/FunctionsTest.php
+++ b/tests/UnitTest/FunctionsTest.php
@@ -152,7 +152,7 @@ class FunctionsTest extends TestCase
         $this->assertTrue($definition instanceof EnvironmentVariableDefinition);
         $this->assertEquals('foo', $definition->getVariableName());
         $this->assertTrue($definition->isOptional());
-        $this->assertSame(null, $definition->getDefaultValue());
+        $this->assertNull($definition->getDefaultValue());
     }
 
     /**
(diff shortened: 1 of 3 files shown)


SvenRtbg:
Good thing!

How's anyones thought about `assertEquals` (compares with `==`) vs. `assertSame` (compares with `===`). I do prefer the latter and try to avoid the former if at all possible because the same rule also applies to the production code.

tests/UnitTest/FunctionsTest.php
```diff
@@ -152,7 +152,7 @@ public function test_env_default_value_null()
         $this->assertTrue($definition instanceof EnvironmentVariableDefinition);
         $this->assertEquals('foo', $definition->getVariableName());
         $this->assertTrue($definition->isOptional());
-        $this->assertSame(null, $definition->getDefaultValue());
```
mnapoli: I'm not sure `assertSame()` should be replaced here, it's not exactly the same thing. IIRC `assertNull` does a loose comparison (`== null`) which isn't the same as a strict one (`assertSame`).

carusogabriel: @mnapoli Actually no, is a strict comparison. See [this PHPUnit method](https://github.com/sebastianbergmann/phpunit/blob/master/src/Framework/Constraint/IsNull.php#L27) that checks this conditional.

mnapoli: Awesome TIL

Thanks!

PR #569, opened 2018-01-19
Fix #567 Ignore errors for invalid referenced definitions during compilation

### LINKED ISSUE #567: Recursive compilation breaks when abstract classes are involved
As first mentioned here https://github.com/PHP-DI/PHP-DI/pull/566#issuecomment-357530850, it "can" occur that an abstract class is involved in the collection of classes that is considered for compilation.

Simply skipping these classes might be an approach to be considered?

**Debugging / logging**
As mentioned in https://github.com/PHP-DI/PHP-DI/pull/566#issuecomment-357691451 it is sometimes difficult to determine "why" such an abstract class is considered for compilation.

Improving / extending the error message, or the ability to inject a PSR Logger might simplify this.

mnapoli:
Fixes #567

### DIFF
--- a/tests/IntegrationTest/CompiledContainerTest.php
+++ b/tests/IntegrationTest/CompiledContainerTest.php
@@ -7,6 +7,7 @@ namespace DI\Test\IntegrationTest;
 use function DI\autowire;
 use DI\ContainerBuilder;
 use function DI\create;
+use DI\Definition\Exception\InvalidDefinition;
 use function DI\get;
 
 /**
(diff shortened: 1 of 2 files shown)


tests/IntegrationTest/CompiledContainerTest.php
```diff
@@ -171,6 +172,36 @@ public function recursively_compiles_referenced_definitions_found()
...
+    public function invalid_definitions_referenced_in_the_configuration_throw_an_error()
+    {
+        $builder = new ContainerBuilder;
+        $builder->addDefinitions([
+            CompiledContainerTest\ConstructorWithAbstractClassTypehint::class => autowire(),
+            CompiledContainerTest\AbstractClass::class => autowire(),
+        ]);
+        $builder->enableCompilation(self::COMPILATION_DIR, self::generateCompiledClassName());
+        $builder->build();
+    }
```
amakhrov: The test name says `throw an error` - but there is no `@expectedException` annotation. Since the test passed, I guess no error is actually triggered in this case.
Am I just missing the point?

mnapoli: Oh wait… You are right I iterated several times on the solution and in the end it's not working as expected, thanks for catching that!

mnapoli: It should be good now.

PR #572, opened 2018-01-24
Fix issues found by static analysis

muglug:
Ran Psalm (latest dev-master) on it using [this psalm.xml config](https://gist.github.com/muglug/6671b531ca249637d3d3d303a3632221) and added 7.0-safe return types automatically.

### DIFF
--- a/src/CompiledContainer.php
+++ b/src/CompiledContainer.php
@@ -92,8 +92,6 @@ abstract class CompiledContainer extends Container
 
     /**
      * Invoke the given callable.
-     *
-     * @return mixed
      */
     protected function resolveFactory($callable, $entryName, array $extraParameters = [])
     {
--- a/src/CompiledContainer.php
+++ b/src/CompiledContainer.php
@@ -92,6 +92,8 @@ abstract class CompiledContainer extends Container
 
     /**
      * Invoke the given callable.
+     *
+     * @return mixed
      */
     protected function resolveFactory($callable, $entryName, array $extraParameters = [])
     {
--- a/src/Compiler/Compiler.php
+++ b/src/Compiler/Compiler.php
@@ -298,7 +298,7 @@ PHP;
     }
 
     /**
-     * @return string|null|true If null is returned that means that the value is compilable.
+     * @return string|true If null is returned that means that the value is compilable.
      */
     private function isCompilable($value)
     {
--- a/src/Compiler/Compiler.php
+++ b/src/Compiler/Compiler.php
@@ -256,7 +256,7 @@ PHP;
         // Check that the value can be compiled
         $errorMessage = $this->isCompilable($value);
         if ($errorMessage !== true) {
-            throw new InvalidDefinition($errorMessage);
+            throw new InvalidDefinition((string) $errorMessage);
         }
 
         if ($value instanceof Definition) {
@@ -298,7 +298,7 @@ PHP;
     }
 
     /**
-     * @return string|null If null is returned that means that the value is compilable.
+     * @return string|null|true If null is returned that means that the value is compilable.
      */
     private function isCompilable($value)
     {
--- a/src/Definition/Resolver/ArrayResolver.php
+++ b/src/Definition/Resolver/ArrayResolver.php
@@ -58,9 +58,6 @@ class ArrayResolver implements DefinitionResolver
         return true;
     }
 
-    /**
-     * @return mixed
-     */
     private function resolveDefinition(Definition $value, ArrayDefinition $definition, $key)
     {
         try {
--- a/src/Definition/Resolver/ArrayResolver.php
+++ b/src/Definition/Resolver/ArrayResolver.php
@@ -35,11 +35,11 @@ class ArrayResolver implements DefinitionResolver
      *
      * An array definition can contain simple values or references to other entries.
      *
-     * @param ArrayDefinition $definition
-     *
      * {@inheritdoc}
+     *
+     * @param ArrayDefinition $definition
      */
-    public function resolve(Definition $definition, array $parameters = [])
+    public function resolve(Definition $definition, array $parameters = []): array
     {
         $values = $definition->getValues();
 
--- a/src/Definition/Resolver/ArrayResolver.php
+++ b/src/Definition/Resolver/ArrayResolver.php
@@ -58,6 +58,9 @@ class ArrayResolver implements DefinitionResolver
         return true;
     }
 
+    /**
+     * @return mixed
+     */
     private function resolveDefinition(Definition $value, ArrayDefinition $definition, $key)
     {
         try {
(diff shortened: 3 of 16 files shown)


src/Compiler/Compiler.php
```diff
@@ -298,7 +298,7 @@ private function createCompilationDirectory(string $directory)
     }
 
     /**
-     * @return string|null If null is returned that means that the value is compilable.
+     * @return string|null|true If null is returned that means that the value is compilable.
```
mnapoli: We can probably replace the `|null` by `|true` entirely? (there is no `return null`)

src/Definition/Resolver/ResolverDispatcher.php
```diff
@@ -29,11 +29,34 @@ class ResolverDispatcher implements DefinitionResolver
...
     private $objectResolver;
+
+    /**
+     * @var InstanceInjector|null
+     */
     private $instanceResolver;
+
+    /**
+     * @var EnvironmentVariableResolver|null
+     */
```
mnapoli: I'd like to avoid this documentation as it's not helping in this specific class.

src/Definition/Source/DefinitionArray.php
```diff
@@ -28,7 +28,7 @@ class DefinitionArray implements DefinitionSource, MutableDefinitionSource
 
     /**
      * Cache of wildcard definitions.
-     * @var array
+     * @var array|null
```
mnapoli: That's not really the case as the array is assigned in the constructor.

src/Definition/Source/SourceChain.php
```diff
@@ -41,8 +41,11 @@ public function __construct(array $sources)
 
     /**
      * {@inheritdoc}
+     *
      * @param int $startIndex Use this parameter to start looking from a specific
      *                        point in the source chain.
+     *
+     * @return null|Definition
```
mnapoli: This is covered by the parent method, it's not necessary here.

src/Definition/Source/AnnotationBasedAutowiring.php
```diff
@@ -52,6 +52,9 @@ public function __construct($ignorePhpDocErrors = false)
         $this->ignorePhpDocErrors = (bool) $ignorePhpDocErrors;
     }
 
+    /**
+     * @return ObjectDefinition|null
+     */
```
mnapoli: This is covered by the parent method, it's not necessary here.

src/Definition/Source/DefinitionArray.php
```diff
@@ -80,6 +80,9 @@ public function addDefinition(Definition $definition)
         $this->wildcardDefinitions = null;
     }
 
+    /**
+     * @return null|Definition
+     */
```
mnapoli: This is covered by the parent method, it's not necessary here.

src/Definition/Source/ReflectionBasedAutowiring.php
```diff
@@ -15,6 +15,9 @@
  */
 class ReflectionBasedAutowiring implements DefinitionSource, Autowiring
 {
+    /**
+     * @return ObjectDefinition|null
+     */
```
mnapoli: This is covered by the parent method, it's not necessary here.

src/Compiler/Compiler.php
muglug: Will change

src/Definition/Source/AnnotationBasedAutowiring.php
muglug: You mean the interface? I'll remove it

Psalm doesn't inherit docs from interfaces, because two conflicting interface definitions cause problems, but I could make a general exception when they're only defined in one.

src/Definition/Source/DefinitionArray.php
muglug: Yes, but set to `null` in a couple of places, and checked for null in the `getDefinition` method

src/Definition/Source/SourceChain.php
muglug: Will remove

src/Definition/Source/ReflectionBasedAutowiring.php
muglug: Will remove

src/Definition/Source/DefinitionArray.php
muglug: Will remove

src/Definition/Resolver/ResolverDispatcher.php
muglug: I'll remove, but I'm curious why you want less documentation – are this class's properties in flux?

mnapoli: They are not really helpful (really they bring no value to understand the class, on the contrary they add noise)

src/Definition/Source/DefinitionArray.php
mnapoli: Good point thanks!

src/Definition/Resolver/ResolverDispatcher.php
muglug: I've removed them.

Would you feel differently if the property types were supported by the language (so you could write `private ?SomeResolver $someResolver`)?

mnapoli: Definitely. It's just that this specific class is a dispatcher and has many properties, and those are all very explicit. The annotation was redundant.

src/CompiledContainer.php
```diff
@@ -92,6 +92,8 @@ protected function setDefinition(string $name, Definition $definition)
 
     /**
      * Invoke the given callable.
+     *
+     * @return mixed
```
mnapoli: This is not necessary, especially in a protected function.

src/Definition/Resolver/ResolverDispatcher.php
```diff
@@ -30,10 +30,15 @@ class ResolverDispatcher implements DefinitionResolver
...
     private $arrayResolver;
+
     private $factoryResolver;
+
     private $decoratorResolver;
+
     private $objectResolver;
+
     private $instanceResolver;
+
```
mnapoli: Could you simply revert this file? (there are still changes here)

src/Definition/Resolver/ArrayResolver.php
```diff
@@ -58,6 +58,9 @@ public function isResolvable(Definition $definition, array $parameters = []) : b
         return true;
     }
 
+    /**
+     * @return mixed
+     */
```
mnapoli: Not necessary, especially on a private method.

src/CompiledContainer.php
muglug: I'm removed these. You have a bunch of other `@return mixed` on protected/private methods, assumed you wouldn't mind more of them.

mnapoli:
Thanks!

Issue #604, opened 2018-06-11
Make CompiledContainer idempotent?


holtkamp:
Heroku [offers free "dyno's" / containers, which go to "sleep" after idling 30 minutes](https://devcenter.heroku.com/articles/free-dyno-hours), in the logs this looks like this:

```
2018-06-08T13:36:14.928015+00:00 heroku[web.1]: Idling
2018-06-08T13:36:14.928583+00:00 heroku[web.1]: State changed from up to down
2018-06-08T13:36:15.659288+00:00 heroku[web.1]: Stopping all processes with SIGTERM
2018-06-08T13:36:15.772020+00:00 heroku[web.1]: Process exited with status 143
2018-06-08T13:36:15.671374+00:00 app[web.1]: Going down, terminating child processes...
2018-06-11T06:38:43.485395+00:00 heroku[web.1]: Unidling
2018-06-11T06:38:43.485781+00:00 heroku[web.1]: State changed from down to starting
2018-06-11T06:38:52.247525+00:00 heroku[web.1]: Starting process with command `vendor/bin/heroku-php-apache2 public/`
2018-06-11T06:38:56.892979+00:00 app[web.1]: DOCUMENT_ROOT changed to 'public/'
2018-06-11T06:38:56.910706+00:00 app[web.1]: Using WEB_CONCURRENCY=2 processes.
2018-06-11T06:38:56.918291+00:00 app[web.1]: Starting php-fpm...
2018-06-11T06:38:58.919763+00:00 app[web.1]: Starting httpd...
2018-06-11T06:38:59.300606+00:00 heroku[web.1]: State changed from starting to up
```

After waking up, a problem seems to occurs with the combination of:

- PHP-DI CompiledContainer
- [OPCache file cache](http://php.net/manual/en/book.opcache.php)
  - Note that [OPCache file cache](http://php.net/manual/en/opcache.configuration.php#ini.opcache.file-cache) is enabled: OPCache entries are also saved to disk, so upon server start they are available immediately. Not sure what parts of this Heroku "keeps" between idling / unindling...

```
Array
(
    [type] => 1
    [message] => Uncaught Error: Call to undefined method CompiledContainer::get5b1e190a4fff9261052747() in /app/vendor/php-di/php-di/src/CompiledContainer.php:53
Stack trace:
#0 /app/data/cache/CompiledContainer.php(39133): DI\CompiledContainer->get('settings.amazon...')
[...]

PR #605, opened 2018-06-12
Use hashed values for method names in the CompiledContainer

holtkamp:
As suggested in https://github.com/PHP-DI/PHP-DI/issues/604

### DIFF
--- a/src/Compiler/Compiler.php
+++ b/src/Compiler/Compiler.php
@@ -157,6 +157,17 @@ class Compiler
     private function compileDefinition(string $entryName, Definition $definition) : string
     {
         $methodName = $this->getHashedValue('get', $entryName);
+
+        //In case an Entry is already added, the used method should be equal
+        if(isset($this->entryToMethodMapping[$entryName]) && $this->entryToMethodMapping[$entryName] !== $methodName){
+            throw new InvalidDefinition(sprintf(
+                'Entry "%s" cannot be compiled. An Entry with the same name already exists pointing to method %s(), while this one points to method %s().',
+                $entryName,
+                $this->entryToMethodMapping[$entryName],
+                $methodName
+            ));
+        }
+
         $this->entryToMethodMapping[$entryName] = $methodName;
 
         switch (true) {
--- a/src/Compiler/Compiler.php
+++ b/src/Compiler/Compiler.php
@@ -16,6 +16,7 @@ use DI\Definition\Source\DefinitionSource;
 use DI\Definition\StringDefinition;
 use DI\Definition\ValueDefinition;
 use DI\DependencyException;
+use DI\Proxy\ProxyFactory;
 use InvalidArgumentException;
 use PhpParser\Node\Expr\Closure;
 use SuperClosure\Analyzer\AstAnalyzer;
@@ -63,6 +64,21 @@ class Compiler
      * @var bool
      */
     private $autowiringEnabled;
+
+    /**
+     * @var ProxyFactory
+     */
+    private $proxyFactory;
+
+    public function __construct(ProxyFactory $proxyFactory)
+    {
+        $this->proxyFactory = $proxyFactory;
+    }
+
+    public function getProxyFactory() : ProxyFactory
+    {
+        return $this->proxyFactory;
+    }
 
     /**
      * Compile the container.
--- a/src/Compiler/Compiler.php
+++ b/src/Compiler/Compiler.php
@@ -143,15 +143,10 @@ class Compiler
 
     /**
      * Use a hash to ensure that the used method names in the CompiledContainer are both unique and idempotent.
-     *
-     * @param string $prefix
-     * @param string $value
-     *
-     * @return string
      */
     private function getHashedValue(string $prefix, string $value): string
     {
-        return sprintf('%s%s', $prefix, md5($value));
+        return $prefix . md5($value);
     }
 
     /**
@@ -161,7 +156,7 @@ class Compiler
      */
     private function compileDefinition(string $entryName, Definition $definition) : string
     {
-        $methodName = $this->getHashedValue('get', $entryName . $definition);
+        $methodName = $this->getHashedValue('get', $entryName);
         $this->entryToMethodMapping[$entryName] = $methodName;
 
         switch (true) {
--- a/src/Compiler/Compiler.php
+++ b/src/Compiler/Compiler.php
@@ -142,14 +142,26 @@ class Compiler
     }
 
     /**
+     * Use a hash to ensure that the used method names in the CompiledContainer are both unique and idempotent.
+     *
+     * @param string $prefix
+     * @param string $value
+     *
+     * @return string
+     */
+    private function getHashedValue(string $prefix, string $value): string
+    {
+        return sprintf('%s%s', $prefix, md5($value));
+    }
+
+    /**
      * @throws DependencyException
      * @throws InvalidDefinition
      * @return string The method name
      */
     private function compileDefinition(string $entryName, Definition $definition) : string
     {
-        // Generate a unique method name
-        $methodName = str_replace('.', '', uniqid('get', true));
+        $methodName = $this->getHashedValue('get', $entryName . $definition);
         $this->entryToMethodMapping[$entryName] = $methodName;
 
         switch (true) {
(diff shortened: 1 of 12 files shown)


src/Compiler/Compiler.php
```diff
@@ -260,8 +272,8 @@ public function compileValue($value) : string
         }
 
         if ($value instanceof Definition) {
-            // Give it an arbitrary unique name
-            $subEntryName = uniqid('SubEntry');
+            $subEntryName = $this->getHashedValue('SubEntry', $value->getName() . $value);
```
mnapoli: I am cautious about this part: nested definitions don't have names, so `getName()` will probably be null here.

And casting `$value` to string dumps the debug output of the definition. That output is not unique ([for example](https://github.com/PHP-DI/PHP-DI/blob/fd7bde0fbd25b7e4ca913907d274b9a71278e347/src/Definition/FactoryDefinition.php#L77)).

I'm not sure what to do here that would not require bigger changes 🤔

tests/IntegrationTest/CompiledContainerTest.php
```diff
@@ -9,6 +9,8 @@
 use function DI\create;
 use DI\Definition\Exception\InvalidDefinition;
 use function DI\get;
+use DI\Test\IntegrationTest\Definitions\NestedDefinitionsTest\AllKindsOfInjections;
+use DI\Test\IntegrationTest\Definitions\NestedDefinitionsTest\Autowireable;
```
mnapoli: Could you copy those classes in this file (like [done here](https://github.com/PHP-DI/PHP-DI/blob/79712dc23ae1a86824dde1ee659ddc45449d275c/tests/IntegrationTest/Definitions/NestedDefinitionsTest.php#L238-L265))? I refactored the tests with v6 so that most integration tests are isolated from each other, that is much easier to maintain.

src/Compiler/Compiler.php
```diff
@@ -141,15 +141,27 @@ public function compile(
...
+     * Use a hash to ensure that the used method names in the CompiledContainer are both unique and idempotent.
+     *
+     * @param string $prefix
+     * @param string $value
+     *
+     * @return string
+     */
+    private function getHashedValue(string $prefix, string $value): string
+    {
+        return sprintf('%s%s', $prefix, md5($value));
```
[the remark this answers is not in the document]
holtkamp: Haha, guess that is my personal quirk of loving `sprintf()`

src/Compiler/Compiler.php
holtkamp: yeah, was getting that concern this morning too. But then again: SubEntries get the value of there "parent" entry appended as well since the `$subEntryName` is handed over to `compileDefinition()`... and the parent must be unique, right?

Maybe some more exhaustive testing could be added. For what its worth: adopted this fork in production without problems

tests/IntegrationTest/CompiledContainerTest.php
holtkamp: sure, was in a hurry

holtkamp:
Thanks for the feedback, will try to process it soon!

src/Compiler/Compiler.php
```diff
@@ -141,15 +141,27 @@ public function compile(
...
     /**
      * @throws DependencyException
      * @throws InvalidDefinition
      * @return string The method name
      */
     private function compileDefinition(string $entryName, Definition $definition) : string
     {
-        // Generate a unique method name
-        $methodName = str_replace('.', '', uniqid('get', true));
+        $methodName = $this->getHashedValue('get', $entryName . $definition);
```
mnapoli: Oh I missed the concatenation to the `$definition` cast to string. I don't think that is necessary as entry names are unique (and that is expensive since `$definition` is an object).

src/Compiler/Compiler.php
mnapoli: I'm not sure I understand what you mean. Here the `$subEntryName` is used to create a method in the compiled container class, it needs to be unique in itself?

Maybe what we could do is propagate the parent entry names down the nested definitions somehow (that way we can prefix those). But I'm afraid that means adding extra parameters everywhere.

src/Compiler/Compiler.php
holtkamp: confirmed, appending the string representation of the Definition is only required for the subEntries

holtkamp:
@mnapoli processed your feedback and pushed the changes. 

Can you think of some additional Definitions to add to the tests? As indicated: in a real-life situation no collisions occurred, so it seems it all works out. But it is probably better to be safe than sorry so the more we test, the better 😄

src/Compiler/Compiler.php
holtkamp: >I'm not sure I understand what you mean.

Ah, yeah sorry, that did not make sense. 

So theoretically a collision could occur in case multiple Definitions exist which have:
- the name "<nested definition>", for example for a
  - `DI\Definition\Reference`
  - `DI\Definition\ArrayDefinition`
  - ... ?? more possible?
- the same value

Let's add a (failing) test for that first

holtkamp:
I added some tests. For now I conclude that when a collision occurs in the entryName resulting from the hash, the value of the Definition should be the same. 

Added a check for that. In case they ARE the same (which they should), then this can be considered an optimization: the amount of functions with different names, but equal body (so functionally nothing changes), is reduced.

src/Compiler/Compiler.php
mnapoli: That is interesting! FYI I won't have time to look further into that in the next days. We could add a new method on definitions (e.g. `getHash()`) if necessary.

Also I think we need to check explicitly for collisions with closures, because `__toString()` is very poor right now so I think there may be collisions.

holtkamp: >I won't have time to look further into that in the next days.

ok, standing by, for now this approach resolves our issue. 

>We could add a new method on definitions (e.g. getHash()) if necessary.

Indeed, but then a Definition needs to be aware that it is a SubEntry as well?

>Also I think we need to check explicitly for collisions with closures, because __toString() is very poor right now so I think there may be collisions.

True, my knowledge of the internals is too limited to set up proper tests...

Step-by-step we'll get there 😄

holtkamp:
FYI: rebased changes on master branch

holtkamp:
FYI: rebased changes on master branch

@mnapoli I know you are busy, but do you, or someone else, have ideas on getting this functionality merged?

mnapoli:
@holtkamp really sorry about the delay. Regarding the last discussion left open in the diff it seems to me you've covered it with a check (an exception is thrown). Is that right? That seems like a good call, at least a clear exception is thrown. (just want to make sure before I merge, it's been some time 😄)

By the way have you been able to use this branch in production?

holtkamp:
@mnapoli indeed, a `InvalidDefinition` Exception is thrown in case "equal hashes are involved, which do not point to the same method name". 

>By the way have you been able to use this branch in production?

yes, since mid-june 2018, works properly / no strange side-effects...

src/Compiler/Compiler.php
```diff
@@ -157,15 +157,33 @@ public function compile(
         return $fileName;
     }
 
+    /**
+     * Use a hash to ensure that the used method names in the CompiledContainer are both unique and idempotent.
+     */
+    private function getHashedValue(string $prefix, string $value): string
+    {
+        return $prefix . md5($value);
```
elie29: @holtkamp IMHO, I think we should not use md5 but sha1 or hash function

holtkamp: @elie29 ok, but why would that have your preference? Due to the chance of collision? That change is really small..

https://stackoverflow.com/questions/201705/how-many-random-elements-before-md5-produces-collisions

Or is it for performance reasons?

mnapoli: A quick search shows that md5 is faster than sha1 so that sounds better to use md5.

elie29: @holtkamp Just due to collision

mnapoli: I really don't think we have to worry about that given the size of the sample.

mnapoli:
Thank you!

I've tagged a new release: https://github.com/PHP-DI/PHP-DI/releases/tag/6.0.7

holtkamp:
:shipit:

PR #607, opened 2018-06-23
Added instructions of using PHP-DI with Zend Expressive

elie29:
[zend-phpdi-config](https://github.com/elie29/zend-di-config) acts as a bridge for zend expressive projects. It was merged in zend expressive skeleton.


couscous.yml
```diff
@@ -67,6 +67,12 @@ menu:
                 zf1:
                     text: Zend Framework 1
                     url: doc/frameworks/zf1.html
+                ze:
+                    text: Zend Expressive
+                    absoluteUrl: https://github.com/elie29/zend-di-config
+                slim:
+                    text: Slim
+                    url: doc/frameworks/slim.html
```
mnapoli: It seems you have duplicated the `slim` entry here

couscous.yml
```diff
@@ -67,6 +67,12 @@ menu:
                 zf1:
                     text: Zend Framework 1
                     url: doc/frameworks/zf1.html
+                ze:
+                    text: Zend Expressive
+                    absoluteUrl: https://github.com/elie29/zend-di-config
```
mnapoli: Since you have created a page for that (which is perfect) you can change this link to point to:

```yaml
url: doc/frameworks/zend-expressive.html
```

And by the way could you rename the `ze.md` file to `zend-expressive.md`? Some people browse the documentation on GitHub so I think it would be better to have more explicit file names.

You can keep `ze` for the menu entry though that's fine, it doesn't appear anywhere.

elie29: thank you @mnapoli I'll do the modification

mnapoli:
Awesome thank you!

PR #611, opened 2018-07-30
Use fqn to helpers in example code

ferdypruis:
Updated some examples to use the fully qualified name to the helper functions, as most examples already did.


mnapoli:
Thanks!

PR #616, opened 2018-08-25
Adds support for autowire of __construct for wildcards

jasonrm:
In 5.x it was possible to have auto wiring of wildcard definitions via reflection, this PR enables that functionality again.

The "fix" might be a bit hacky however, as I wasn't able to find a cleaner way for ReflectionBasedAutowiring to be aware of what class to use reflection on, unless I sent the wildcard class through the normalizer a second time after the derived implementation class name was set.

### DIFF
--- a/src/Definition/Source/DefinitionArray.php
+++ b/src/Definition/Source/DefinitionArray.php
@@ -115,6 +115,7 @@ class DefinitionArray implements DefinitionSource, MutableDefinitionSource
                     $definition->setClassName(
                         $this->replaceWildcards($definition->getClassName(), $matches)
                     );
+                    $definition = $this->normalizer->normalizeRootDefinition($definition, $name);
                 }
 
                 return $definition;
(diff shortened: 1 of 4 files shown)


mnapoli:
Thanks for the PR, you have actually spotted a bug! Wildcard definitions with autowiring are supposed to work in 6.x, and they work in simple scenarios. Your PR fixes the bug which is great.

I think however there is a more elegant solution because your solution involves "normalizing" the definition twice. I've opened a PR on your PR in your repository ;) -> https://github.com/ArtFireCom/PHP-DI/pull/1 If you merge it into your branch then this pull request here will be updated and good to merge I think.

jasonrm:
Excellent! This was exactly the best case scenario I'd hoped for where,
1) it wasn't me using the library incorrectly
2) with your deep understanding of the code you could provide an even better solution that didn't resort to a hacky fix

mnapoli:
Great. And woops it seems I missed an unused import (see https://github.com/PHP-DI/PHP-DI/pull/616/checks?check_run_id=13995093) would you have time to remove it? That way the build will be green and I'll be able to merge and release.

PR #617, opened 2018-08-27
Remove Container::get() InvalidArgumentException

b1rdex:
Long time ago there was `is_string` check of `$name`. I can't find commit that removed it.
Personally I'd rather return `is_string` check, becuase currently `array_key_exists` will emit warning if `$name` is not a string and not an int.
If you decide to return `is_string` check, than `InvalidArgumentException` should be replaced with `NotFoundException` or `DependencyException` to not break interface contract.


mnapoli:
Perfect thanks!

Issue #614, opened 2018-08-16
Interface parameters are not correctly autowired


Agares:
This works correctly on PHP-DI 5, but not on 6:
```php
<?php
require_once __DIR__.'/vendor/autoload.php';
interface A {}
class B implements A {}

class C {
	function __construct(A $a) {}
}
$builder = new DI\ContainerBuilder();
$builder->useAnnotations(false);
$builder->useAutowiring(true);
$container = $builder->build();
$container->set(A::class, \DI\get(B::class));

$c = $container->get(C::class);

var_dump($c);
```

Is this change of behaviour intended? I haven't found anything about this in the migration guide.

Output of the script on PHP-DI6:
```
$ php test.php
PHP Fatal error:  Uncaught DI\Definition\Exception\InvalidDefinition: Entry "C" cannot be resolved: Entry "A" cannot be resolved: the class is not instantiable
Full definition:
Object (
    class = #NOT INSTANTIABLE# A
    lazy = false
)
Full definition:
Object (
    class = C
    lazy = false
    __construct(
        $a = get(A)
    )
) in C:\Users\[REDACTED]\Desktop\dupa\vendor\php-di\php-di\src\Definition\Exception\InvalidDefinition.php:18
Stack trace:
#0 C:\Users\[REDACTED]\Desktop\dupa\vendor\php-di\php-di\src\Definition\Resolver\ObjectCreator.php(155): DI\Definition\Exception\InvalidDefinition::create(Object(DI\Definition\ObjectDefinition), 'Entry "C" canno...')
#1 C:\Users\[REDACTED]\Desktop\dupa\vendor\php-di\php-di\src\Definition\Resolver\ObjectCreator.php(71): DI\Definition\Resolver\ObjectCreator->createInstance(Object(DI\Definition\ObjectDefinition), Array)
#2 C:\Users\[REDACTED]\Desktop\dupa\vendor\php-di\php-di\src\Definition\Resolver\ResolverDispatcher.php(64): DI\Definition\Resolver\ObjectCr in C:\Users\[REDACTED]\Desktop\dupa\vendor\php-di\php-di\src\Definition\Exception\InvalidDefinition.php on line 18
```

`composer show`:
```
$ composer show
jeremeamia/SuperClosure 2.4.0  Serialize Closure objects, including their context and binding
nikic/php-parser        v4.0.3 A PHP parser written in PHP
php-di/invoker          2.0.0  Generic and extensible callable invoker
[...]

PR #621, opened 2018-08-31
Test to make sure container can set and resolve interface - issue 641

smytsyk:
DI\get was able to resolve binding on interface in PHP-DI 5, however it's not working on v6 anymore.

https://github.com/PHP-DI/PHP-DI/issues/614


mnapoli:
Thanks, I've opened #622 with a fix.

PR #624, opened 2018-09-06
Add support for variadic arguments

### LINKED ISSUE #619: Support for variadic arguments
PHP-DI does not currently support variadic arguments for autowiring. It will only inject the first element to the variadic parameter in a constructor, silently discarding the rest.

For example for this class:

```PHP
class UserProfile implements Timeline\Engine
{
    public function __construct(DbConnection $dbConnection, TimelineModule ...$modules)
    {
```

cannot be autowired by:

```PHP
            Engine::class => \DI\autowire(UserProfile::class)
                ->constructor(
                    \DI\get(DbConnection::class),
                    \DI\get(Module\CasinoFollow::class),
                    \DI\get(Module\UserFollow::class),
                    \DI\get(Module\Review::class)
                ),
```

but must be instead autowired by:

```PHP
            UserProfile::class => function(ContainerInterface $c) {
                return new UserProfile(
                    $c->get(DbConnection::class),
                    $c->get(Module\CasinoFollow::class),
                    $c->get(Module\UserFollow::class),
                    $c->get(Module\Review::class)
                );
            },
```

which in turn reduces the efficiency of caching I believe and is harder to debug.

tssge:
This closes #619 

This pull request is probably missing a lot of things, as I am not completely familiar with PHP-DI source yet. Please feel free to comment on what I could improve.

These changes pass all tests and a new test was added to support the new behavior.


mnapoli:
You seem to have errors in code formatting (https://github.com/PHP-DI/PHP-DI/pull/624/checks). Could you fix them? And you will probably need to merge master in your branch because of an update in Symfony's coding style.

tssge:
Thank you for your feedback, I have fixed the formatting now

tests/IntegrationTest/Definitions/AutowireDefinitionTest.php
```diff
@@ -343,6 +344,30 @@ public function test_php71_nullable_typehint(ContainerBuilder $builder)
...
+     */
+    public function test_all_variadic_arguments_are_passed(ContainerBuilder $builder)
+    {
+        $arguments = ['test1', 'test2', 'test3'];
+
+        $container = $builder
+            ->useAutowiring(true)
+            ->addDefinitions([
+                Variadic::class => autowire()
+                    ->constructor(...$arguments)
```
mnapoli: Could you write the same test for a *setter* method too? (I want to make sure that the patch covers both constructors and methods)

src/Definition/Resolver/ParameterResolver.php
```diff
@@ -48,7 +48,7 @@ public function resolveParameters(
 
         $definitionParameters = $definition ? $definition->getParameters() : [];
 
-        foreach ($method->getParameters() as $index => $parameter) {
+        foreach ($this->getMethodParameters($method, count($definitionParameters)) as $index => $parameter) {
```
mnapoli: I'll check what it means performance-wise but it seems the new method call doesn't do much so it should be negligible 👍

tests/IntegrationTest/Definitions/AutowireDefinitionTest.php
tssge: I have added a test for method injection

src/Definition/Resolver/ParameterResolver.php
tssge: Indeed, I found the function call much more clear to whoever is reading the code, but if it seems to make performance worse it could be inlined

zelgerj:
Hello, i would be interested about the status of this, since I also faced the same issue with variadic arguments. Thanks

Issue #618, opened 2018-08-28
[QueryBuilder] Allow to provide an "index" of Definitions?


holtkamp:
When a lot of DIC Definitions are involved, I found it was useful to organize them in separate files, which contain arrays:

```php
//Config1.php
return [
        LoggerInterface::class => \DI\create(Logger::class),
        //...
];
```

Then a single "index" file can be used to refer the these separate files:

```php
//definitionsFiles.php
return [
    __DIR__ . \DIRECTORY_SEPARATOR . 'Config1.php',
    __DIR__ . \DIRECTORY_SEPARATOR . 'Config2.php',
    __DIR__ . \DIRECTORY_SEPARATOR . 'ConfigN.php',
];
```
Then this "index" of Definitions can be processed like this:

```php
$definitionsFiles = require 'definitionsFiles.php';
foreach($definitionsFiles as $definitionsFile{
    $containerBuilder->addDefinitions($definitionsFile);
}
```

Would a PR with a function that accepts such an "index" of Definitions be accepted, or is out of of scope? Something like this:

```php
//QueryBuilder.php
public function addDefinitionsIndex(string $definitionsIndex): self{
    foreach($definitionsIndex as $definitions{
        $containerBuilder->addDefinitions($definitions);
    }
}
```

mnapoli:
I would rather like to keep the API as small as possible ("don't make me think"). Here we would just avoid a foreach so there doesn't seem to be a lot of added value.

I wonder though if `addDefinitions()` could be changed to take a list of files/arrays using variadic parameters. That way the API stays the same but it can support lists of definitions. Example of usage:

```php
$builder->addDefinitions(...$definitionsIndex);
```

What do you think?

holtkamp:
Aah, yeah, variadic parameters would be a great alternative to reach the same goal, nice!

juliangut:
IMO the way you store your definitions and how you collect them should be out of the container

If you know you're not overriding any definition (DI\add or DI\decorate) you can just do this:

```
$builder->addDefinitions(\array_merge(...$definitionsIndex));
```

holtkamp:
@juliangut note that in my example the `$definitionsIndex` is an array of strings (pointing to the configuration files) and not an array of arrays.

mnapoli:
> IMO the way you store your definitions and how you collect them should be out of the container

Agreed. Although if the variadic things work out that is a very little impact on the API (almost invisible) and easy to maintain so that's why I think "why not".

[...]

PR #627, opened 2018-09-17
support variadic parameter on addDefinitions

juliangut:
As per this comments: https://github.com/PHP-DI/PHP-DI/issues/618#issuecomment-416686620 https://github.com/PHP-DI/PHP-DI/issues/618#issuecomment-416954644

Support variadic arguments when calling addDefinitions

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -45,5 +45,6 @@ return PhpCsFixer\Config::create()
         'strict_comparison' => true,
         'strict_param' => true,
         'yoda_style' => false,
+        'native_function_invocation' => false,
     ])
     ->setFinder($finder);
--- a/src/ContainerBuilder.php
+++ b/src/ContainerBuilder.php
@@ -306,9 +306,9 @@ class ContainerBuilder
     /**
      * Add definitions to the container.
      *
-     * @param string|array|DefinitionSource $definitions Can be an array of definitions, the
-     *                                                   name of a file containing definitions
-     *                                                   or a DefinitionSource object.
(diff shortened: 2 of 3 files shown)


juliangut:
The build error is coming from PHPStan

```
------ ------------------------------------------------------------------
 Line src/ContainerBuilder.php
 ------ ------------------------------------------------------------------
  314    PHPDoc tag @param for parameter $definitions with type
array<array|DI\Definition\Source\DefinitionSource|string> is not
subtype of native type array<int, mixed>
 ------ ------------------------------------------------------------------
```

This is the code

```php
    /**
     * Add definitions to the container.
     *
     * @param string[]|array[]|DefinitionSource[] $definitions Can be an array of definitions, the
     *                                                   name of a file containing definitions
     *                                                   or a DefinitionSource object.
     * @return $this
     */
    public function addDefinitions(...$definitions) : self
    {
    [...]
```

mnapoli:
Regarding the coding style I've added a commit on master to fix the coding style (Symfony added a new rule that breaks BC): a9892481aec265b83591a444f1a33ceaf6b44cfb If you merge/rebase master your PR should be OK.

As for phpstan I think you don't need to update the types in the phpdoc, try this and see if it works:

```
string|array|DefinitionSource
```

juliangut:
Finally! dockblock syntax on variadic arguments does not seam to be well documented

https://github.com/phpstan/phpstan/issues/979#issuecomment-386222212

mnapoli:
Thanks!

PR #630, opened 2018-10-03
Simplify conditions

omarkdev:
This PR is intended to simplify conditions, either by applying an early return or apply ternary operations in `if` and `else` conditions

### DIFF
--- a/src/Definition/Resolver/ParameterResolver.php
+++ b/src/Definition/Resolver/ParameterResolver.php
@@ -71,9 +71,11 @@ class ParameterResolver
 
             // Nested definitions
             if ($value instanceof Definition) {
-                $value = $parameter->isOptional() && ! $this->definitionResolver->isResolvable($value)
-                    ? $this->getParameterDefaultValue($parameter, $method)
-                    : $this->definitionResolver->resolve($value);
+                if ($parameter->isOptional() && ! $this->definitionResolver->isResolvable($value)) {
+                    $value = $this->getParameterDefaultValue($parameter, $method);
+                } else {
+                    $value = $this->definitionResolver->resolve($value);
+                }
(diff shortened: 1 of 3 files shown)


src/Definition/Resolver/ParameterResolver.php
```diff
@@ -71,12 +71,9 @@ public function resolveParameters(
...
             if ($value instanceof Definition) {
-                // If the container cannot produce the entry, we can use the default parameter value
-                if ($parameter->isOptional() && ! $this->definitionResolver->isResolvable($value)) {
-                    $value = $this->getParameterDefaultValue($parameter, $method);
-                } else {
-                    $value = $this->definitionResolver->resolve($value);
-                }
+                $value = $parameter->isOptional() && ! $this->definitionResolver->isResolvable($value)
+                    ? $this->getParameterDefaultValue($parameter, $method)
+                    : $this->definitionResolver->resolve($value);
```
mnapoli: While the rest are great improvements I don't feel this one is more readable with the change, could you revert this?

omarkdev:
Done, @mnapoli,  I reversed the change in the file

src/Definition/Resolver/ParameterResolver.php
```diff
@@ -71,7 +71,6 @@ public function resolveParameters(
 
             // Nested definitions
             if ($value instanceof Definition) {
-                // If the container cannot produce the entry, we can use the default parameter value
```
mnapoli: I just notice this removal: why remove the comment?

omarkdev:
@mnapoli I removed the comments by mistake, now I revert the remove of comment

mnapoli:
Thanks

PR #631, opened 2018-10-07
Set allow risky in php_cs

omarkdev:
Recently I needed to run `php-cs-fixer fix` and I receive the error message showing that I need to set the `--allow-risky` flag, and to solve this confusion in the future I add this flag in the `.php_cs` file


mnapoli:
:+1:

PR #632, opened 2018-10-08
Set correctly phpDocs from Container

omarkdev:
Looking in `Container` class I noticed the class has some errors phpDocs annotations, this PR has the intention to solve this errors.

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -120,6 +120,7 @@ class Container implements ContainerInterface, FactoryInterface, InvokerInterfac
      *
      * @throws DependencyException Error while resolving the entry.
      * @throws NotFoundException No entry found for the given name.
+     * @throws InvalidDefinition
      * @return mixed
      */
     public function get($name)
@@ -145,6 +146,7 @@ class Container implements ContainerInterface, FactoryInterface, InvokerInterfac
      * @param string $name
      *
      * @return Definition|null
+     * @throws InvalidDefinition
      */
     private function getDefinition($name)
     {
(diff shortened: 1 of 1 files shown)


src/Container.php
```diff
@@ -120,6 +120,7 @@ public function __construct(
      *
      * @throws DependencyException Error while resolving the entry.
      * @throws NotFoundException No entry found for the given name.
+     * @throws InvalidDefinition
```
mnapoli: Where is this exception thrown in the method?

src/Container.php
```diff
@@ -230,7 +234,7 @@ public function has($name)
      *
      * @param object $instance Object to perform injection upon
      * @throws InvalidArgumentException
-     * @throws DependencyException Error while injecting dependencies
+     * @throws InvalidDefinition
```
mnapoli: Could you explain this change?

src/Container.php
```diff
@@ -311,6 +318,7 @@ public function getKnownEntryNames() : array
      *
      * @param string $name Entry name
      *
+     * @return string
```
mnapoli: This duplicates information that already exists in the code, please remove it as well as similar changes below.

src/Container.php
```diff
@@ -172,6 +174,7 @@ private function getDefinition($name)
      * @throws InvalidArgumentException The name parameter must be of type string.
      * @throws DependencyException Error while resolving the entry.
      * @throws NotFoundException No entry found for the given name.
+     * @throws InvalidDefinition
```
tssge: This exception bubbles up to Container::get() and should be wrapped by an exception implementing ContainerExceptionInterface in order to stay compliant with the SHOULD clause provided by PSR-11

Though Container::get() is not strictly PSR-11 compliant currently either because it can throw InvalidArgumentException

src/Container.php
```diff
@@ -202,6 +205,7 @@ public function make($name, array $parameters = [])
      * @param string $name Entry name or a class name.
      *
      * @throws InvalidArgumentException The name parameter must be of type string.
+     * @throws InvalidDefinition
```
tssge: You should add InvalidDefinition to the PHPDoc of FactoryInterface::make() as well

PR #642, opened 2019-01-05
Adds return values to docblock

bjoern-tantau:
Adds the return values of the functions to the docblock to aid IDEs.

### DIFF
--- a/src/functions.php
+++ b/src/functions.php
@@ -18,6 +18,8 @@ if (! function_exists('DI\value')) {
      * Helper for defining a value.
      *
      * @param mixed $value
+     *
+     * @return ValueDefinition
      */
     function value($value) : ValueDefinition
     {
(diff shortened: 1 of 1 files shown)


mnapoli:
These docblocks are duplicating information which is already in the code (the return types). I don't think this helps IDEs at all since the information is already there.

bjoern-tantau:
Ah, you're right. My IDE (Netbeans) wasn't showing any autocompletion for these method's return-values and I thought it was because of the missing docblocks, but I guess it must be some bug with Netbeans, as it doesn't even work with the docblock, but on other functions. Maybe because of the namespace.

mnapoli:
Good to know, thanks for the follow up and sorry :)

bjoern-tantau:
For anyone with the same problem, I opened a bug report for Netbeans at https://issues.apache.org/jira/browse/NETBEANS-1859

Issue #542, opened 2017-11-05
Pre-generate proxy classes


mnapoli:
Proxy classes are generated on demand (in dev and production).

It could be useful to pre-generate them for performance reasons and to make it usable on read-only filesystems.

bbaga:
If no one else is working on this issue, I'd be happy to pick it up.

SvenRtbg:
Doesn't this look like it should be closed? Haven't looked in detail, but the referenced PR looks like it addresses this issue, and it got merged.

mnapoli:
:+1:

PR #645, opened 2019-02-15
Adding support for proxy class generation on container compilations

bbaga:
We've run into an issue where we would have to write custom workaround to pre-generate all proxy classes in our container and it isn't very pretty. Would be much nicer if this functionality would be provided out of the box.

This is my attempt generate the proxies when we build the container.

I wonder if it would be beneficial to put the proxy classes into the container file, like Symfony DI does it.

Issue: #542


tests/IntegrationTest/CompiledContainerTest.php
```diff
@@ -157,6 +157,26 @@ public function the_compiled_container_can_extend_a_custom_class()
         self::assertInstanceOf(CompiledContainerTest\CustomParentContainer::class, $container);
     }
 
+    /**
+     * @test
+     */
+    public function compilation_generates_proxy_classes()
```
mnapoli: `proxy_classes_can_be_pregenerated_at_compile_time` ?

doc/lazy-injection.md
```diff
@@ -140,3 +140,15 @@ $containerBuilder->writeProxiesToFile(true, __DIR__ . '/tmp/proxies');
...
+
+By default the proxies are written to disk the first time they are required. Enabling pre-generation will write the proxy classes to disk when the container is built.
+
+```php
+// Enable writing proxies to file in the tmp/proxies directory on contaier build
+$containerBuilder->writeProxiesToFile(true, __DIR__ . '/tmp/proxies');
+$containerBuilder->pregenerateProxiesToFile(true);
+``` 
+
+For this functionality to work, both configuration options have to be set.
```
mnapoli: We should explain also that the container must be compiled for this to work?

src/Compiler/Compiler.php
```diff
@@ -64,6 +65,28 @@ class Compiler
...
 
+    /**
+     * @var ProxyFactory
+     */
+    private $proxyFactory;
+
+    /**
+     * Compiler constructor.
+     * @param ProxyFactory $proxyFactory
+     */
```
mnapoli: This comment can be removed as it duplicates the types in the code.

src/Compiler/Compiler.php
```diff
@@ -64,6 +65,28 @@ class Compiler
...
+    public function __construct(ProxyFactory $proxyFactory)
+    {
+        $this->proxyFactory = $proxyFactory;
+    }
+
+    /**
+     * @return ProxyFactory
+     */
```
mnapoli: Same here the comment duplicates the code.

src/Compiler/ObjectCreationCompiler.php
```diff
@@ -137,16 +137,17 @@ private function compileLazyDefinition(ObjectDefinition $definition) : string
...
+        '$object = $this->proxyFactory->createProxy(
+            ' . $definition->getClassName() . ',
+            function (&$wrappedObject, $proxy, $method, $params, &$initializer) {
+                $wrappedObject = ' . $subDefinition . ';
+                $initializer = null; // turning off further lazy initialization
                 return true;
             }
-        );
-PHP;
+        );';
```
mnapoli: Could you restore the formatting, I see no reason to change this as this is the exact same code.

doc/lazy-injection.md
```diff
@@ -140,3 +140,15 @@ $containerBuilder->writeProxiesToFile(true, __DIR__ . '/tmp/proxies');
...
 You will need to clear the directory every time you deploy to avoid keeping outdated proxies.
+
+### Generating the proxy classes on container build time
+
+By default the proxies are written to disk the first time they are required. Enabling pre-generation will write the proxy classes to disk when the container is built.
+
+```php
+// Enable writing proxies to file in the tmp/proxies directory on contaier build
+$containerBuilder->writeProxiesToFile(true, __DIR__ . '/tmp/proxies');
+$containerBuilder->pregenerateProxiesToFile(true);
```
mnapoli: I'm thinking: do we even need this option at all? If we know all the lazy classes, I see no reason why we shouldn't generate all proxies automatically at build time.

That would make the feature simpler as it would remove the feature flag 🎉 

Do you see a case where we wouldn't want proxies to be generated?

src/Compiler/ObjectCreationCompiler.php
bbaga: Sure, I can do that. The reason for this change is that the formatting was a little off, like:

```php
    protected function get5c68983987f32630765897()
    {
                $object = $this->proxyFactory->createProxy(
            'stdClass',
            function (&$wrappedObject, $proxy, $method, $params, &$initializer) {
                $wrappedObject = $this->get5c689839880df096916618();
                $initializer = null; // turning off further lazy initialization
                return true;
            }
        );
        return $object;
    }
```

and this is because we write the first 8 spaces with

```php
       return <<<PHP
        \$object = \$this->proxyFactory->createProxy(
```

although this is only an "issue" when somebody looks at the compiled container.

doc/lazy-injection.md
bbaga: I was on the edge with this one. I'm not sure how would you want to release this change, but it might surprise some people when the proxy classes appear earlier then they'd expect them and so I chose to with an opt-in flag.

Personally I don't have anything making this the default behaviour, it is up to you.

mnapoli: I thought about it and I don't see any disadvantage. I think it also make sense since we compile the container, so we expect files to be generated. And if those files are ignored when deploying to production they will be regenerated anyway.

The worse case scenario is there is an unforeseen problem because of that, in which case I'll re-add the option then.

So :+1: to remove the option and make it the default behavior when compiling.

src/Compiler/ObjectCreationCompiler.php
mnapoli: Right I see, we have the same problem on other parts of the compiled container, at the moment formatting that compiled file is not a priority.

doc/lazy-injection.md
bbaga: Changed the example and the wording a little

src/Proxy/LazyLoadingValueHolderFactory.php
```diff
@@ -0,0 +1,17 @@
...
+
+declare(strict_types=1);
+
+namespace DI\Proxy;
+
+/**
+ * To expose the least amount of the ProxyManager's internals we are extending the LazyLoadingValueHolderFactory
+ * The protected `generateProxy` method will generate AND write the proxy classes to disk.
+ */
+class LazyLoadingValueHolderFactory extends \ProxyManager\Factory\LazyLoadingValueHolderFactory
```
mnapoli: Waiting on a clarification in https://github.com/Ocramius/ProxyManager/issues/440 if we can avoid this.

mnapoli:
@bbaga I sent a pull request on your pull request 😄 : https://github.com/bbaga/PHP-DI/pull/1

The goal was to simplify the changes by removing the custom class that extends ProxyManager. What I did is I invoke the proxy factory instead, we just don't use the created instance. At least we don't deal with ProxyManager internals because I would rather avoid having to update it when new versions are published.

If you merge https://github.com/bbaga/PHP-DI/pull/1 we can then merge this one and I can tag a release.

bbaga:
@mnapoli makes sense, thank you for following up on this! PR merged on my side.

mnapoli:
Thank you!

PR #646, opened 2019-03-01
fix injected container self is a new instance

coderbiq:
fix injected self container is a new instance.

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -106,7 +106,7 @@ class Container implements ContainerInterface, FactoryInterface, InvokerInterfac
 
         // Auto-register the container
         $this->resolvedEntries = [
-            self::class => $this,
+            static::class => $this,
             ContainerInterface::class => $this->delegateContainer,
             FactoryInterface::class => $this,
             InvokerInterface::class => $this,


mnapoli:
This PR needs more explanations (why this change?) and tests.

I'm afraid as well that this could produce unwanted results with the compiled container (or actually any other subclass) as the `DI\Container` entry will no longer be registered.

elie29:
@coderbiq do you extend the container?

coderbiq:
@elie29 yes, i extend the Di\Container. case is:

```php
class MyContainer extends \DI\Container
{
}
class Foo
{
    public function __construct(MyContainer $c1, \Psr\Container\ContainerInterface $c2)
    {
    }
}

$container = new MyContainer;
$foo       = $container->get(Foo::class);
```

the $c1 is a new MyContainer instance. I Know, use `Psr\Container\ContainerInterface` injected current MyContaienr instance. but the name  Unable to express my design intentions.

my english is not good, so sorry.

elie29:
@coderbiq I don't know if everybody needs or extends the container, however an immediate solution is to have a construct function and add your Container to the resolved entries as follow:

```php
class MyContainer extends \DI\Container
{
    public function __construct(
        MutableDefinitionSource $definitionSource = null,
        ProxyFactory $proxyFactory = null,
        ContainerInterface $wrapperContainer = null
    ) {
       parent::__construct($definitionSource, $proxyFactory, $wrapperContainer);

        // Auto-register current container
        $this->resolvedEntries = [
            static::class => $this,
        ];
    }
}
```

mnapoli:
👍 	for @elie29's suggestion, it's a good way to go. Changing this in PHP-DI itself will break the `DI\Container` entry.

Issue #589, opened 2018-03-29
Allow custom cache implementation


brstgt:
Currently, the cache implementation is hard-wired. I'd like to replace the apcu cache with an in-memory cache.

Why?
- I don't want to roll out apc just for that
- I don't want a persistent cache
- I don't want a cache that shares a namespace among processes (if not configurable)
- I simply don't need it. We have  long running processes (amphp/aerys) and a in-mem cache is totally sufficient

Implementing a cache is super ease but I can't plug it in. To make it work I had to copy + adapt the whole container-builder :(

How a bout ContainerBuilder::setCache($cacheObj) or ::setCacheImplementation($className)?

mnapoli:
To answer your first question:

> Why?

http://php-di.org/doc/performances.html#caching

> The cache relies on APCu directly because it is the only cache system that makes sense (very fast to write and read). Other caches are not good options, this is why PHP-DI does not use PSR-6 or PSR-16 for this cache.

Using Redis/Memcached for that would not make sense as it would be too slow.

In your use case (aerys) I understand an array cache (in-memory) could make sense though. Do you have performance issues if you disable APCu?

Re-introducing an abstraction for the cache could be doable, but I'd rather avoid that if I can because it's a pain to maintain, document and explain.

brstgt:
Yes we did have immense performance penalties. An array-cache worked fine for us.
You don't need an abstraction IMHO. You could simply use existing interfaces. Actually it's only a `DefinitionSource` decorating all other sources, right?

brstgt:
Otherwise add a simple option to use apcu or array-cache. An array cache is only a few lines different from apc cache

mnapoli:
> Otherwise add a simple option to use apcu or array-cache

;) For specific feature requests I can work on commission if your company is willing to support the project. Else I will consider this issue (and the solutions) based on priority (number of people/projects impacted, interest/criticity for the project, etc.) and how much personal time I can spend :)

brstgt:
I can offer a PR but I won't spend time on that if it's not clear that it will be accepted.

PR #647, opened 2019-03-01
WIP: Pluggable caching backends

bzikarsky:
# Overview 

This PR relates to #589 and refactors `SourceCache` into pluggable caching-backends while keeping backwards compatibility.

# Current state

- [x] Implemented new caching classes
- [x] Integrated new caching classes into `ContainerBuilder`
- [ ] Added unit-tests for the new classes

# Contents

## Create an explicit, new class hierarchy for definition caching  

- Establish interface `Cache\Cache` which represents any definition-cache
- Move logic from `SourceCache` to both `Cache\AbstractCache` and
  `Cache\ApcuCache`
- Deprecate `SourceCache` in favour of `Cache\ApcuCache` but still
  keep its functionality by extending the former. This is to keep
  backwards compatibility
- Create a new simple cache-backend `ArrayCache` which just caches
  lookups in memory

## Integrate new Cache backend into ContainerBuilder 

Instead of tracking the caching status with a boolean, this commit changes
`$sourceCache` to a nullable string. This allows to define different cache-
backends.

`::enableDefinitionCache(..)` not takes an optional parameter - optional to
keep backwards compatibility - with the class-name of the definition cache to
use. By default the `ApcuCache` is used which is identical to the deprecated,
original `SourceCache`

## Tests

TODO :smile: 

# Future work

1) This functionality can be used to provide adapters to more general caching-backends, such as PSR-6 or PSR-16 compatible libraries.

2) In a major release `SourceCache` can be removed.

### DIFF
--- a/src/ContainerBuilder.php
+++ b/src/ContainerBuilder.php
@@ -6,16 +6,18 @@ namespace DI;
 
 use DI\Compiler\Compiler;
 use DI\Definition\Source\AnnotationBasedAutowiring;
+use DI\Definition\Source\Cache\ApcuCache;
+use DI\Definition\Source\Cache\Cache;
 use DI\Definition\Source\DefinitionArray;
 use DI\Definition\Source\DefinitionFile;
 use DI\Definition\Source\DefinitionSource;
 use DI\Definition\Source\NoAutowiring;
 use DI\Definition\Source\ReflectionBasedAutowiring;
-use DI\Definition\Source\SourceCache;
 use DI\Definition\Source\SourceChain;
 use DI\Proxy\ProxyFactory;
 use InvalidArgumentException;
 use Psr\Container\ContainerInterface;
+use RuntimeException;
 
 /**
  * Helper to create and configure a Container.
@@ -96,9 +98,9 @@ class ContainerBuilder
     private $compileToDirectory;
 
     /**
-     * @var bool
+     * @var string|null
      */
-    private $sourceCache = false;
+    private $sourceCache;
 
     /**
      * Build a container configured for the dev environment.
@@ -150,12 +152,18 @@ class ContainerBuilder
         // Mutable definition source
         $source->setMutableDefinitionSource(new DefinitionArray([], $autowiring));
 
-        if ($this->sourceCache) {
-            if (!SourceCache::isSupported()) {
-                throw new \Exception('APCu is not enabled, PHP-DI cannot use it as a cache');
+        if ($this->sourceCache !== null) {
+            $isSupported = [$this->sourceCache, 'isSupported'];
+            $createInstance = [$this->sourceCache, 'create'];
+
+            if (!$isSupported()) {
+                throw new RuntimeException(sprintf(
+                    'Cache backend "%s" is not supported, PHP-DI cannot use it as a cache',
+                    $this->sourceCache
+                ));
             }
             // Wrap the source with the cache decorator
-            $source = new SourceCache($source);
+            $source = $createInstance($source);
         }
 
         $proxyFactory = new ProxyFactory(
@@ -350,13 +358,20 @@ class ContainerBuilder
      *
      * @see http://php-di.org/doc/performances.html
      *
+     * @param string $cacheBackend
      * @return $this
      */
-    public function enableDefinitionCache() : self
+    public function enableDefinitionCache(string $cacheBackend = ApcuCache::class) : self
     {
         $this->ensureNotLocked();
 
-        $this->sourceCache = true;
+        if (!in_array(Cache::class, class_implements($cacheBackend), true)) {
+            throw new InvalidArgumentException(
+                'The passed cache-backend hs to implement the ' . Cache::class . ' interface'
+            );
+        }
+
+        $this->sourceCache = $cacheBackend;
 
         return $this;
     }
--- a/src/Definition/Source/DefinitionSource.php
+++ b/src/Definition/Source/DefinitionSource.php
@@ -17,6 +17,7 @@ interface DefinitionSource
     /**
      * Returns the DI definition for the entry name.
      *
+     * @param string $name The name of the definition to get
      * @throws InvalidDefinition An invalid definition was found.
      * @return Definition|null
      */
--- a/src/Definition/Source/SourceCache.php
+++ b/src/Definition/Source/SourceCache.php
@@ -2,76 +2,20 @@
 
 namespace DI\Definition\Source;
 
-use DI\Definition\AutowireDefinition;
-use DI\Definition\Definition;
-use DI\Definition\ObjectDefinition;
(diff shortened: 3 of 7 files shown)


mnapoli:
Hi, thank you for the effort you put into that!

Unfortunately I'm still not sure I understand the need behind that.

The other thing is that this adds a lot of complexity and new code to maintain, and using an abstraction or PSR-6/16 adds overhead as well. I did some testing back then and using Doctrine Cache (which is fairly optimized) did add some overhead compared to a custom and simple cache implementation around APCu directly.

src/Definition/Source/SourceCache.php
```diff
@@ -2,76 +2,20 @@
...
+    public function __construct(DefinitionSource $source)
     {
-        return
-            // Cache missing definitions
-            ($definition === null)
-            // Object definitions are used with `make()`
-            || ($definition instanceof ObjectDefinition)
-            // Autowired definitions cannot be all compiled and are used with `make()`
-            || ($definition instanceof AutowireDefinition);
+        parent::__construct($source);
```
SvenRtbg: If this parent-call is all you do in the constructor, you can remove the constructor entirely.

src/Definition/Source/DefinitionSource.php
```diff
@@ -17,6 +17,7 @@ interface DefinitionSource
     /**
      * Returns the DI definition for the entry name.
      *
+     * @param string $name The name of the definition to get
```
SvenRtbg: Not a big fan of adding obvious annotations. The type is already enforced, and the explanation for a method "getDefinition" with the parameter "name" is also pretty obvious: You get the definition for the name.

src/ContainerBuilder.php
```diff
@@ -96,9 +98,9 @@ class ContainerBuilder
     private $compileToDirectory;
 
     /**
-     * @var bool
+     * @var string|null
      */
-    private $sourceCache = false;
+    private $sourceCache;
```
SvenRtbg: I'd probably use another name for this property for defining the class name instead of recycling a property that already has a meaning, especially if it's type would change.

src/Definition/Source/SourceCache.php
bzikarsky: Visibility-change for BC! (original method is `protected`)

src/Definition/Source/DefinitionSource.php
bzikarsky: I think if there is a docblock, it should be complete, shouldn't it? I'm all for dropping the comment after `$name` though.

src/ContainerBuilder.php
bzikarsky: Renaming the property is an a non-issue. What would you prefer?

bzikarsky:
The need for this is simple: If you run PHP in any long-running-process configuration (php-pm, amp, etc.) it's fully sufficient to simply cache in-memory. APCu is an overhead for us, we don't even run it in production. I also think that the introduced overhead of 2 method calls is negligable. In the end it's still only 1 object which does the same as `SourceCache` - only that `apc_fetch` and `apc_store` are moved behind 2 methods.

In the end it comes down to this: We have currently a copy of `SourceCache` and `ContainerBuilder` in our source-tree solely for the purpose to cache into arrays. It'd be great to be able to drop those.

If you give your :+1: I will addthe missing  tests for the new classes.

src/Definition/Source/DefinitionSource.php
SvenRtbg: My personal opinion: Docblocks should be as short as possible, and should not contain redundant information. Comments will start to contain wrong information in the long run, the less info they contain, the better.

Now, PHP-DI is a little bit inconsistent in this regard. If you look over some code, you'll find methods like this https://github.com/PHP-DI/PHP-DI/blob/f16f27b7bc0876c7e76f568e0733d59997c8ec9c/src/ContainerBuilder.php#L226 without @param for the parameter and the opposite like this https://github.com/PHP-DI/PHP-DI/blob/f16f27b7bc0876c7e76f568e0733d59997c8ec9c/src/ContainerBuilder.php#L278 with every parameter annotated and commented.

I cannot define the code style for this project, @mnapoli is the project lead.

src/ContainerBuilder.php
SvenRtbg: `cacheClass`

bzikarsky:
@mnapoli Please consider merging something along these lines. The copied and modified version of ContainerBuilder we need to maintain in our source-tree broke prod builds after a composer upgrade again. 

I'm more than happy to implement the missing tests/refactorings if you greenlight this implementation.

mnapoli:
@bzikarsky sorry for the delay.

The use case makes sense (not using APCu yet caching definitions in memory).

I've been thinking about this and I wonder if it wouldn't make sense to always have the in-memory cache, even when using APCu. Indeed if a process gets the same definition twice it doesn't make sense to fetch it twice from APCu. That wouldn't mean a noticeable performance improvement but I think it might be a simpler architecture (no pluggable cache backend).

What do you think?

bzikarsky:
I agree on having an in-memory cache "by default".

What I don't understand yet is why you insist to not move the "concern to cache" into a separate interface. From an architecture point-of-view smaller classes with a smaller footprint make sense in my opinion (and are also easier to test).

Going with the proposed way would open the code up to all different kind of caching use-cases and with `Memory(APCu())` only being one/the default.

mnapoli:
Well I was thinking of a simple key-value map of definitions that were already resolved but it seems I forgot that this exists already:

https://github.com/PHP-DI/PHP-DI/blob/2c7a3a6b7b6c5dbe299a7a0cde386cea5d095eb3/src/Container.php#L151-L154

Does that cover your use case?

bzikarsky:
So... I know this is a very old PR, but priorities moved, but here we are again.

I checked your proposal, but the issue is:

While `Container::getDefinition` uses the `$this->fetchedDefinitions` cache. `Container::injectOn` does not. We introduced hack which led to this PR, because we have *a lot of* `injectOn` callsites for legacy reasons. 

I am not sure: Maybe it's just possible to make `injectOn` use `fetchedDefinitions` (or `getDefinition`), but I assume there is a reason that `injectOn` neither stores nor reads from this "cache".

I'd like to have your input:
1) Do you think `injectOn` could be modified in such a way this PR is not neccessary anymore?
2) If this is not the case, would you merge this PR if I rebase it on the current master?

mnapoli:
I had a look, I think the `fetchedDefinitions` cache was omitted from `injectOn()` by mistake. I can't recall any reason not to use it there. (the tests should also validate that nothing break if you try to use it)

I think 1. is the simplest solution. If you manage to implement it that would be best!

bzikarsky:
Closed in favour of https://github.com/PHP-DI/PHP-DI/pull/709

PR #652, opened 2019-03-15
Alternative solution to fix #648 and keep the container idempotent

### LINKED ISSUE #648: Error compiling factory definitions since 6.0.7
```php
<?php

declare(strict_types=1);

use function DI\autowire;

use DI\ContainerBuilder;
use function DI\factory;

require_once __DIR__.'/../vendor/autoload.php';

class Fu
{
    public function create(): self
    {
        return new static();
    }
}

class Bar
{
    public function create(): self
    {
        return new static();
    }
}

class FuBar
{
    private $fu;

    private $bar;

    public function __construct(Fu $fu, Bar $bar)
    {
        $this->fu = $fu;
        $this->bar = $bar;
    }
}

$builder->enableCompilation(__DIR__);

$builder->addDefinitions([
    FuBar::class => autowire()
        ->constructorParameter('fu', factory([Fu::class, 'create']))
        ->constructorParameter('bar', factory([Bar::class, 'create']))
]);

$dic = $builder->build();
```
In `CompiledContainer.php` you'll find 

```php
    protected function get3b62a97e28ef909b96b42a793be8a616()
    {
        $object = new FuBar($this->get5ffee756f39d463eb2a0c22906c0f2be(), $this->get5ffee756f39d463eb2a0c22906c0f2be());
        return $object;
    }
```
As you can see the definition for `Bar` (Method `get5ffee756f39d463eb2a0c22906c0f2be`) is used for every Parameter in `FuBar::__construct`.

I was able to nail the issue down to `\DI\Compiler\Compiler::getHashedValue`. In this scenario the parameter `string $value` is `<nested definition>Factory` for *every* factory definition used like above.

In our case this is a crititcal issue, downgrading to 6.0.6 fixed this.

Relates to #605

mnapoli:
This PR builds on top of #649 and implements what's suggested in https://github.com/PHP-DI/PHP-DI/pull/649/files#r264150572

It adds more complexity, but at least we have a sane way to make sure every definition can be hashed to something unique (I am afraid #649 doesn't cover every case because it only focuses on closures).

The idea is that each definition is compiled into a method which bears a unique name. That unique name is a hash generated from the definition.

Each definition generates its own hash, so that two different definitions have two different hashes. But it's okay if different definitions have the same hash if they are actually equal.

### DIFF
--- a/src/Compiler/Compiler.php
+++ b/src/Compiler/Compiler.php
@@ -174,18 +174,6 @@ class Compiler
     {
         $methodName = $this->getHashedValue('get', $entryName);
 
-        //In case an Entry is already added, the used method should be equal
-        if (isset($this->entryToMethodMapping[$entryName]) && $this->entryToMethodMapping[$entryName] !== $methodName) {
-            throw new InvalidDefinition(sprintf(
-                'Entry "%s" cannot be compiled. An Entry with the same name already exists pointing to method %s(), while this one points to method %s().',
-                $entryName,
-                $this->entryToMethodMapping[$entryName],
-                $methodName
-            ));
-        }
-
-        $this->entryToMethodMapping[$entryName] = $methodName;
-
         switch (true) {
             case $definition instanceof ValueDefinition:
                 $value = $definition->getValue();
@@ -262,7 +250,7 @@ PHP;
                     ));
                 }
 
-                if($value instanceof \Closure){
+                if ($value instanceof \Closure) {
                     $methodName .= '_closure_' . \md5($this->compileClosure($value));
                 }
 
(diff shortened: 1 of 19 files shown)


src/Compiler/Compiler.php
```diff
@@ -262,6 +244,10 @@ private function compileDefinition(string $entryName, Definition $definition) :
                     ));
                 }
 
+                if ($value instanceof \Closure) {
+                    $methodName .= '_closure_' . \md5($this->compileClosure($value));
+                }
```
mnapoli: Mmh I thought I removed the need for any specific code like this, I need to check if I didn't simply forget this here.

mnapoli:
Closing per https://github.com/PHP-DI/PHP-DI/issues/648#issuecomment-478951613

Issue #651, opened 2019-03-14
make() is not using typehints but named parameters instead


dynasource:
Let me first thank you for all the efforts put into this useful library.

About the issue. I dont know if this is intended behavior, but the fact that argument names in constructors are used to match with ```make()```` parameters, is quite error prone. 

Once you change a parameter $name, exceptions are thrown.

Am I missing some setting to force Typehint parameter resolving?

mnapoli:
I'm sorry I don't understand.

dynasource:
Let's say one wants to create class X:
```
class X{
public function __construct(Z $name_1)
}
```
By calling make($id, $params):
```
make(X::class, [
'name_1' => new Z ()
])
```

In this basic scenario, one has to make sure 'name_1' and $name_1 are equally named, which is much more error prone, than when is matched by Class typehint (Class_Z)

dynasource:
Of course names are always unique, being it a safe solution compared to a case in which constructors have multiple arguments of the same type. In most cases (apart from variadic) this is rare and a fallback to the current $name matching solution could be executed. 

To me, matching by Class Typehint is the most obvious default as it is less error prone and also done during autowiring.

SvenRtbg:
You may call it ignorant, but I never really found a need to use the unstandardized `make()`, but always went for the PSR-compliant `get()`. And autowiring of course, keeping manual configuration to a minimum.

Yes, you'd have to pay attention to the parameters names, and all the same problems occur that we also have when using strings to define method names for mocks in tests: Once you refactor, either the tests are going red for no real reason, or worse: They stay green.

Maybe `make()` has no good interface because syntactical elements like variable/parameter names are represented as strings and are likely to not be changed during refactoring. But the same is true for adding class names as strings (if you don't use `::class`). 

My major point however would be that you cannot have the same type twice. For classes, this might be only a minor issue, but what about scalar types, arrays?

mnapoli:
@dynasource if you want to inject a dependency (especially via its typehint) then you will want to either:

- let PHP-DI create `Z` (autowiring)
- configure `Z` in the container configuration and not specify this parameter in `make()`

Parameters in `make()` are only useful for parameters that are not dependencies.

But yeah maybe in your case `get()` is better as well.

dynasource:
I tend to disagree here

[...]

PR #653, opened 2019-03-31
Fixes FactoryResolver#resolve ignoring passed parameters

Aijoona:
FactoryResolver#resolve seems to be ignoring the passed arguments.

### DIFF
--- a/src/Definition/Resolver/FactoryResolver.php
+++ b/src/Definition/Resolver/FactoryResolver.php
@@ -75,7 +75,7 @@ class FactoryResolver implements DefinitionResolver
         try {
             $providedParams = [$this->container, $definition];
             $extraParams = $this->resolveExtraParams($definition->getParameters());
-            $providedParams = array_merge($providedParams, $extraParams);
+            $providedParams = array_merge($providedParams, $extraParams, $parameters);
 
             return $this->invoker->call($callable, $providedParams);
         } catch (NotCallableException $e) {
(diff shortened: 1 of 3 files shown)


mnapoli:
Thanks, I'm not sure I understand the problem. Are you able to add an integration test? That should help to understand.

holtkamp:
Maybe relats to https://github.com/PHP-DI/PHP-DI/issues/651 ?

Aijoona:
For more context.

Somewhere in my code, I got something like this (a glorified cURL wrapper):
```PHP

    /**
     * @param $resource
     * @param array $payload
     * @param array $options
     * @return mixed
     */
    public function get($resource, array $payload = [], array $options = [])
    {
        return $this->factory
            ->make(Get::class, [
                'resource' => $resource,
                'payload' => $payload,
                'options' => $options
            ])
            ->exec();
    }
```

In my integration tests I need to fake the cURL call, so I added a new definition for the Get::class that overrides the cURL request  (the Client class just forwards the definitions to the container builder).
```PHP

    protected function getClientWithFixedGetResponse($httpCode, $response)
    {
        return new Client(null, null, [
            'definitions' => [
                Get::class => function ($resource, $payload = [], $options = []) use ($httpCode, $response) {
                    $handler = new FixedResponseGetRequest($resource);
                    $handler->setCallback(function () use ($response, $httpCode) {
                        return [$httpCode, $response];
                    });

                    return $handler;
                }
            ]
        ]);
    }

```
Now, the current implementation ignores all the parameters passed in the make call (in this case {resource, payload, options}), so the FactoryResolver just passes the "default params": {container, definition}.

In my code this means an exception being thrown because obviously the container is not the url string expected.

mnapoli:
Thank you!

PR #661, opened 2019-05-31
Loosen APCu check for SourceCache

falkenhawk:
Use sourceCache if isSupported check passes, otherwise proceed without cache and do not throw an exception

- so that php-di does not break if there is any problem with apcu - instead of having to wrap ContainerBuilder::build() with try...catch block

_[from ovos/PHP-DI#3]_


mnapoli:
Thank you, I am afraid about this change that some people will think they use APCu, but they actually don't.

Could you explain the use case behind this?

PR #662, opened 2019-05-31
AnnotationBasedAutowiring with additional options for performance tweaks

falkenhawk:
- adds possibility to disable features which are not used
- useAnnotations(boolean) toggle on autowire() helper to enable/disable reading annotations on specific definitions
- default behaviour does not change (backwards compatibility)

Available flags for `$builder->useAnnotations(true, $flags)`:
- `AnnotationBasedAutowiring::ALL` - all annotations enabled (default)
- `AnnotationBasedAutowiring::IMPLICIT` - enable on implicit definitions (not specified in container config)
- `AnnotationBasedAutowiring::EXPLICIT` - enable on all autowire definitions (which are written in DI config) by default
- `AnnotationBasedAutowiring::INJECTABLE` - read @Injectable annotations for classes
- `AnnotationBasedAutowiring::PROPERTIES` - read @Inject annotations for properties
- `AnnotationBasedAutowiring::METHODS` - read @Inject annotations for methods' parameters

We are using the following combination of options:
```php
$builder->useAnnotations(true, AnnotationBasedAutowiring::IMPLICIT | AnnotationBasedAutowiring::PROPERTIES);
```

It that case it means that:
1) Annotations will be read when a class is not explicitly defined in definitions file
2) annotations will be disabled by default for classes defined in definitions file, `autowire()->useAnnotations()` to enable annotations
3) `@Injectable` is skipped for performance
4) `@Inject` on properties will be read only when annotations are enabled for an entry (either on implicit definitions or explicit with `useAnnotations()`)
5) `@Inject` on methods will be skipped for performance - as we don't use them

_[from ovos/PHP-DI#4]_

### DIFF
--- a/src/ContainerBuilder.php
+++ b/src/ContainerBuilder.php
@@ -57,6 +57,11 @@ class ContainerBuilder
     private $useAnnotations = false;
 
     /**
+     * @var int
+     */
+    private $annotationsFlags = 0;
+
+    /**
      * @var bool
      */
     private $ignorePhpDocErrors = false;
@@ -126,7 +131,7 @@ class ContainerBuilder
         $sources = array_reverse($this->definitionSources);
 
         if ($this->useAnnotations) {
-            $autowiring = new AnnotationBasedAutowiring($this->ignorePhpDocErrors);
+            $autowiring = new AnnotationBasedAutowiring($this->ignorePhpDocErrors, $this->annotationsFlags);
             $sources[] = $autowiring;
         } elseif ($this->useAutowiring) {
             $autowiring = new ReflectionBasedAutowiring;
@@ -239,11 +244,12 @@ class ContainerBuilder
      *
      * @return $this
      */
-    public function useAnnotations(bool $bool) : self
+    public function useAnnotations(bool $bool, int $flags = 0) : self
     {
         $this->ensureNotLocked();
 
         $this->useAnnotations = $bool;
+        $this->annotationsFlags = $flags;
 
         return $this;
     }
--- a/src/Definition/AutowireDefinition.php
+++ b/src/Definition/AutowireDefinition.php
@@ -9,4 +9,26 @@ namespace DI\Definition;
  */
 class AutowireDefinition extends ObjectDefinition
 {
+    /**
+     * @var bool|null
+     */
+    protected $useAnnotations;
+
+    /**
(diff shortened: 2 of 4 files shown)


SvenRtbg:
There are no tests for that feature. And you should probably back your claim of performance gain with some benchmark tests as well.

mnapoli:
> And you should probably back your claim of performance gain with some benchmark tests as well.

@SvenRtbg I am the one supposed to do the benchmarks, per our previous discussion with OP.

@falkenhawk thanks for opening the PR, I will review them over the coming days (possibly weeks) ;)

mnapoli:
Hi, I'm coming back on this from #726. First, sorry this isn't merged or review yet.

I think this PR adds too many options, at least without a good reason. My goal is to simplify as much as possible.

Could we maybe identify one or two main scenarios/reasons? The way I usually proceed is to ask why until we have a concrete use case:

> adds possibility to disable features which are not used

Why?

> useAnnotations(boolean) toggle on autowire() helper to enable/disable reading annotations on specific definitions

Why?

Also, regarding performances do you have any estimate about how much we could gain? I want to see if that is actually interesting, or not worth the effort.

falkenhawk:
@mnapoli thanks for the update!
tbh I don't remember exactly the details, but my intention was to shave of unnecessary overhead where possible - we were working on a high-availability content delivery system back then. I was profiling the scripts with kcachegrind and identified that (as expected) reading annotations was a significant performance hit.
We had quite big classes with a large number of properties, some of them could not be refactored to use only constructor injections (due to the required effort), and there were a number of `make()` calls, which could not be cached.

It was also done before your improvements to the php-di cache system.

So since we were using annotations _sometimes_ (did I mention that our case was a legacy system which we introduced php-di on top?) we wanted to just skip processing them when they were unnecessary for our cases.

This PR adds options to adjust the "low-level" mechanisms for those who are interested in fine-tuning the performance.

It's a big shame I haven't written any automatic benchmarks back then 😭 i'm sorry, I know it would be much easier to assess the usefulness of each of these tweaks with benchmarks

and ps. I am also quite a bit out of the loop recently, since 99% stuff I do now is js/ts

mnapoli:
Sorry for not being able to follow up on this one more quickly.

I'll put this one aside for now, it introduces more options that I can maintain. Thanks for all this effort however!

PR #665, opened 2019-05-31
Feature/service locator

falkenhawk:
ServiceLocator
- Serving "lazy" dependencies for classes using ServiceSubscriberInterface.
- A ServiceSubscriber exposes its dependencies via static getSubscribedServices() method.
- A ServiceLocator instance could then be injected into a class via constructor or a property - the instance would be already configured with dependences read from getSubscribedServices(), but the dependences won't be instantiated until first get - that's how "laziness" is introduced
- DI\Definition\Reference checks if it's a ServiceLocator entry by comparing its name with DI\Definition\Reference::$serviceLocatorClass
- Reference definitions are passed with additional parameter - $requestingName which generally points to name of the class which implements ServiceSubscriberInterface - to resolve ServiceLocator for that class
- Suggested as a lightweight alternative for heavyweight proxies from ocramius/proxy-manager

note:
- it's based on changes from PHP-DI/PHP-DI#662 (so commit 59a2216 is included here as well) - since there is second argument passed to `Reference` constructor, to also cover `AnnotationBasedAutowiring::readConstructor()` method, which was introduced in that PR

Usage:
```php
class MyClass implements ServiceSubscriberInterface
{
    /**
     * @var ServiceLocator
     */
    protected $serviceLocator;

    /**
     * @param ServiceLocator $serviceLocator
     */
    public function __construct(ServiceLocator $serviceLocator)
    {
        $this->serviceLocator = $serviceLocator;
    }

    /**
     * Lazy instantiate heavy dependencies on-demand
     */
    public static function getSubscribedServices(): array
    {
        return [
            SomeClass::class,
            'heavyDependency' => HeavyService::class,
        ];
    }

    public function doOperation()
    {
        $someClass = $this->serviceLocator->get(SomeClass::class);
        return $someClass->doSomething();
    }

    public function getSomethingFromThatHeavyDependency()
    {
        // this method may be rarely used, and it might be good idea to skip resolving the dependency every time during instantiation for performance reasons
        return $this->serviceLocator->get('heavyDependency')->getSomething();
    }
}

```

_[from ovos/PHP-DI#7]_

### DIFF
--- a/src/CompiledContainer.php
+++ b/src/CompiledContainer.php
@@ -122,14 +122,15 @@ abstract class CompiledContainer extends Container
     }
 
     /**
-     * Resolve ServiceLocator for given subscriber class (based on \DI\Definition\ServiceLocatorDefinition::resolve)
+     * Resolve ServiceLocator for given subscriber class (based on \DI\Definition\ServiceLocatorDefinition::resolve).
      *
      * @param string $requestingName class name of a subscriber, implementing ServiceSubscriberInterface
      * @param string $repositoryClass ServiceLocatorRepository
      * @return ServiceLocator
      * @throws ServiceSubscriberException
      */
-    protected function resolveServiceLocator($requestingName, $repositoryClass) {
+    protected function resolveServiceLocator($requestingName, $repositoryClass)
+    {
         if (!method_exists($requestingName, 'getSubscribedServices')) {
             throw new ServiceSubscriberException(sprintf('The class %s does not implement ServiceSubscriberInterface.', $requestingName));
         }
--- a/src/CompiledContainer.php
+++ b/src/CompiledContainer.php
@@ -120,4 +120,24 @@ abstract class CompiledContainer extends Container
             throw new InvalidDefinition("Entry \"$entryName\" cannot be resolved: " . $e->getMessage());
         }
     }
+
+    /**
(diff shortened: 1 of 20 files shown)


SvenRtbg:
A service locator? Really? That is SO the anti pattern.

If creating a "heavy" dependency that is rarely used is bothering you, there are a couple of options without using the service locator antipattern:

- Separate that part from the main class and instantiate only when needed
- Make that dependency light, i.e. dont do work in the constructor

If you really feel you need a service locator, you could always implement a simple class that gets the DI container instance passed as a parameter and then only resolves specific calls when really needed. This would also not tie your own code to interface implementations of PHP-DI.

Remember that there is no PSR definition for a service locator interface (I guess there are reasons), so by adding such interfaces, you effectively make it impossible to change the dependency container.

mnapoli:
@SvenRtbg please respect others contributions.

You may be confused but `ContainerInterface` is a service locator interface. PHP-DI can be used as a service locator (and it has to be used like this at some point in an application, nothing is magic). Service locator is a pattern that is defined by the usage.

falkenhawk:
@mnapoli Hi, I have recreated the branch, based on latest master and without changes from closed https://github.com/PHP-DI/PHP-DI/pull/662, so feel free to reopen this PR if you are still interested. I hope the "files changed" would then refresh and point to the new branch of the same name 🤔

PR #675, opened 2019-07-04
#674 Special interpretation of `\DI\value` in `\DI\Container::set`

### LINKED ISSUE #674: How to set a closure as value to compiled container?
PHP-DI 6.0.8

I can `set` any value to the `CompiledContainer`, but `Closure`s are always interpreted as factories and `set` throws `LogicException`. I can't even use the `\DI\value()`

```php
<?php

require_once './vendor/autoload.php';

$builder = new \DI\ContainerBuilder();
$builder->enableCompilation('.');

$container->set('db:host', 'localhost');                           // ok
$container->set('db:factory', function(){ /* ... */ });            // LogicException
$container->set('db:factory', \DI\value(function(){ /* ... */ })); // LogicException

// LogicException: You cannot set a definition at runtime on a compiled container.
// You can either put your definitions in a file, disable compilation or ->set()
// a raw value directly (PHP object, string, int, ...) instead of a PHP-DI definition
```

downace:
This fixes #674

`\DI\value` now can be used in `\DI\Container::set` to set raw values. This is useful when you need to set a `Closure` as raw value because by default `Closure`s are interpreted as factories:

```
$container->set('foo', function() { return 'bar' }));
$container->get('foo'); // returns 'bar'

$container->set('foo', \DI\value(function() { return 'bar' })));
$container->get('foo'); // returns Closure
```

### DIFF
--- a/tests/IntegrationTest/ContainerSetTest.php
+++ b/tests/IntegrationTest/ContainerSetTest.php
@@ -7,6 +7,7 @@ namespace DI\Test\IntegrationTest;
 use DI\ContainerBuilder;
 use function DI\create;
 use function DI\get;
+use function DI\value;
 
 /**
  * Tests the set() method from the container.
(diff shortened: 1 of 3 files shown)


tests/IntegrationTest/ContainerSetTest.php
```diff
@@ -58,6 +59,28 @@ public function array_entries_can_be_overridden_by_values(ContainerBuilder $buil
...
+
+        $foo = 'foo';
+        $bar = new ContainerSetTest\Dummy();
+        $baz = function() {};
+
+        $container->set('foo', value($foo));
+        $container->set('bar', value($bar));
+        $container->set('baz', value($baz));
+
+        $this->assertSame($foo, $container->get('foo'));
```
downace: Maybe these two are redundant? The test is primarily focused on closure usage

mnapoli: It's perfect, might as well cover a little bit more.

mnapoli:
Thank you this is perfect!

PR #686, opened 2019-10-02
Make compiled container (almost) idempotent

### LINKED ISSUE #604: Make CompiledContainer idempotent?

diesse:
Using a progressive counter for mapped methods/sub entries, container recompilation produces always the same result if definitions do not change.
It's not fully idempotent because changing definitions order leads to a different container, even if functionally unchanged.
Partially fixes #604

### DIFF
--- a/src/Compiler/Compiler.php
+++ b/src/Compiler/Compiler.php
@@ -52,12 +52,12 @@ class Compiler
      * @var int
      */
     private $subEntryCounter;
-    
+
     /**
      * @var int
      */
     private $methodMappingCounter;
-    
+
     /**
      * Map of entry names to method names.
      *
--- a/src/Compiler/Compiler.php
+++ b/src/Compiler/Compiler.php
@@ -49,11 +49,23 @@ class Compiler
     private $entriesToCompile;
 
     /**
+     * Progressive counter for definitions.
+     *
+     * Each key in $entriesToCompile is defined as 'SubEntry' + counter
+     * and each definition has always the same key in the CompiledContainer
+     * if PHP-DI configuration does not change
+     *
      * @var int
      */
     private $subEntryCounter;
 
     /**
+     * Progressive counter for CompiledContainer get methods
+     *
+     * Each CompiledContainer method name is defined as 'get' + counter
+     * and remains the same after each recompilation
+     * if PHP-DI configuration does not change
+     *
      * @var int
      */
     private $methodMappingCounter;
@@ -175,7 +187,7 @@ class Compiler
     private function compileDefinition(string $entryName, Definition $definition) : string
     {
         // Generate a unique method name
-        $methodName = 'get' . sprintf('%022d', ++$this->methodMappingCounter);
+        $methodName = 'get' . (++$this->methodMappingCounter);
         $this->entryToMethodMapping[$entryName] = $methodName;
 
         switch (true) {
@@ -287,7 +299,7 @@ PHP;
 
         if ($value instanceof Definition) {
             // Give it an arbitrary unique name
-            $subEntryName = 'subEntry' . sprintf('%013d', ++$this->subEntryCounter);
+            $subEntryName = 'subEntry' . (++$this->subEntryCounter);
             // Compile the sub-definition in another method
             $methodName = $this->compileDefinition($subEntryName, $value);
             // The value is now a method call to that method (which returns the value)
--- a/src/Compiler/Compiler.php
+++ b/src/Compiler/Compiler.php
@@ -53,18 +53,18 @@ class Compiler
      *
      * Each key in $entriesToCompile is defined as 'SubEntry' + counter
      * and each definition has always the same key in the CompiledContainer
-     * if PHP-DI configuration does not change
+     * if PHP-DI configuration does not change.
      *
      * @var int
      */
     private $subEntryCounter;
 
     /**
-     * Progressive counter for CompiledContainer get methods
+     * Progressive counter for CompiledContainer get methods.
      *
      * Each CompiledContainer method name is defined as 'get' + counter
      * and remains the same after each recompilation
-     * if PHP-DI configuration does not change
+     * if PHP-DI configuration does not change.
      *
      * @var int
      */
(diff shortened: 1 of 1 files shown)


holtkamp:
Nice idea 😸

src/Compiler/Compiler.php
```diff
@@ -48,6 +48,16 @@ class Compiler
      */
     private $entriesToCompile;
 
+    /**
+     * @var int
```
mnapoli: Could you add a description to this field to explain how/why it's different from the other one?

src/Compiler/Compiler.php
```diff
@@ -48,6 +48,16 @@ class Compiler
...
+     */
+    private $subEntryCounter;
+
+    /**
+     * @var int
+     */
+    private $methodMappingCounter;
```
mnapoli: Could you add a description to this field to explain how/why it's different from the other one?

src/Compiler/Compiler.php
```diff
@@ -165,7 +175,7 @@ public function compile(
     private function compileDefinition(string $entryName, Definition $definition) : string
     {
         // Generate a unique method name
-        $methodName = str_replace('.', '', uniqid('get', true));
+        $methodName = 'get' . sprintf('%022d', ++$this->methodMappingCounter);
```
mnapoli: Why is it `%022d` here and `%013d` in the other place? I'm not familiar with this syntax, why use this instead of a concatenation?

src/Compiler/Compiler.php
diesse: I added a description for the fields

src/Compiler/Compiler.php
diesse: It's useless, indeed :blush:
The syntax adds 0-padding to the left of the counter in order to keep the same string length of the original naming scheme.

mnapoli:
Thanks @diesse!

Let's give this a go in a new release!

PR #692, opened 2019-10-31
Improve documentation concerning CompositeContainer

holtkamp:
No more "acclimation" is required, since PHP-DI now implements ContainerInterface

https://github.com/AcclimateContainer/acclimate-container/blob/8c3373637900c48aa4a950bdb7099ac881d8f382/src/ContainerAcclimator.php#L68


mnapoli:
Thanks!

Issue #693, opened 2019-11-07
Wrong order of DefaultValueResolver in Container::getInvoker()


kakenbok:
The `DefaultValueResolver` currently gets executed before the `TypeHintContainerResolver`. This causes type hinted parameters with default values always to resolve to the default value. Thus, it's impossible to overload php functions with new parameters, which requires any additional parameter to default to `null` in order to prevent the PHP warning: `Declaration of ... should be compatible with ...`

```
public function call(Service $service = null) {
  // $service is always null
}

$container->set(Service::class, $service);
$container->call([$obj, 'call']);
```
**Fix1**

`Container.php:402` needs to be changed from:

```
$parameterResolver = new ResolverChain([
    ...
    new DefaultValueResolver,
    new TypeHintContainerResolver($this->delegateContainer)
]);
```

to:

```
$parameterResolver = new ResolverChain([
    ...
    new TypeHintContainerResolver($this->delegateContainer),
    new DefaultValueResolver
]);
```
**Or Fix2**

Allow for injecting:
- a custom `ResolverChain`
- or a custom `Invoker` instance or class

kakenbok:
Fix2 already has an issue: Allow for custom parameter resolvers #553

kakenbok:
Pull request with fix + test: #694

mnapoli:
Closed per https://github.com/PHP-DI/PHP-DI/pull/694: this is the intended behavior.

PR #694, opened 2019-11-07
Wrong order of DefaultValueResolver in Container::getInvoker() #693

kakenbok:
PR for Fix1 of #693

### DIFF
--- a/doc/frameworks/zf2.md
+++ b/doc/frameworks/zf2.md
@@ -90,7 +90,7 @@ Annotations are disabled by default since PHP-DI 5. To enable them, use the foll
 ```php
 return [
     'phpdi-zf2' => [
-        'useAnntotations' => true,
+        'useAnnotations' => true,
     ]
 ];
 ```


mnapoli:
Thanks!

I am surprised no test failed 🤔 This behavior is intended as far as I recall.

The logic behind that is that most optional dependencies are so that we can override the default services to customize the behavior. E.g. pass a custom logger instead of the default logger that the service will use.

If PHP-DI provided those services by default, then the default behavior of some services could be surprising (because it's no longer "the default").

I am afraid that merging this will also break many expected behaviors out there.

One way to solve this is to explicitly pass the parameter when calling `$container->call('service name', ['parameter' => $value])`.

Let me know if I misunderstood something.

kakenbok:
Would it be possible to you briefly explaining this case (with some code): 

> we can override the default services to customize the behavior. E.g. pass a custom logger instead of the default logger that the service will use.

I really tried hard, but i don't get it yet. :-)

PR #696, opened 2019-12-03
Add 6.0 migration link

SvenRtbg:
I hope this makes it automatically to php-di.org once merged


mnapoli:
Good catch thanks!

mnapoli:
It's online: http://php-di.org/doc/migration/6.0.html !

PR #697, opened 2019-12-12
enable apcu on travis + php 7.3

falkenhawk:
also skip reporting unmatched "Function apcu_.* not found." errors from phpstan thus failing travis build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ php:
   - 7.0
   - 7.1
   - 7.2
+  - 7.3
   - nightly
 
 matrix:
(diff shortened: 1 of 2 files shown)


mnapoli:
Thanks!

falkenhawk:
@mnapoli you're welcome :) Btw. I haven't added php 7.4 yet, despite php-di alone being 100% compatible, but tests throw deprecation errors. And actually not php-di tests, but phpunit-easymock:
image

mnapoli:
Oh that's interesting, thanks for looking into that!

PR #698, opened 2019-12-12
SourceCache: namespacing cache keys

falkenhawk:
Prevent cache collisions when sharing a single APC memory pool between multiple DI containers


mnapoli:
Thanks!

PR #699, opened 2019-12-12
Enable PHP 7.4 on travis

jdreesen:
PHP 7.4 is latest and should be tested, too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ php:
   - 7.1
   - 7.2
   - 7.3
+  - 7.4
   - nightly
 
 matrix:
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ before_script:
 script:
   - if [[ $(phpenv version-name) == '7.2' ]]; then vendor/bin/phpunit --coverage-clover clover.xml ; fi
   - if [[ $(phpenv version-name) != '7.2' ]]; then vendor/bin/phpunit ; fi
-  - if [[ $(phpenv version-name) != '7.4' ]]; then vendor/bin/phpstan analyse -l 6 -c phpstan.neon src ; fi
+  - if [[ $(phpenv version-name) == '7.2' ]]; then vendor/bin/phpstan analyse -l 6 -c phpstan.neon src ; fi
 
 after_script:
   - if [[ $(phpenv version-name) == '7.2' ]]; then php vendor/bin/coveralls -v ; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,8 +7,6 @@ notifications:
     on_success: never
 
 php:
-  - 7.0
-  - 7.1
   - 7.2
   - 7.3
   - 7.4
@@ -19,7 +17,7 @@ matrix:
   allow_failures:
     - php: nightly
   include:
-    - php: 7.0
+    - php: 7.2
       env: dependencies=lowest
 
 cache:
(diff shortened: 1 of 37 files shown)


falkenhawk:
@jdreesen of course! The reason I haven't added it to #697 was that it's failing, as I explained in https://github.com/PHP-DI/PHP-DI/pull/697#issuecomment-564909915 - https://github.com/mnapoli/phpunit-easymock has to be adjusted first

jdreesen:
Yes, I just read your comment in the other PR. I'm not sure the problem is easymock, though, but rather PHPUnit itself.

jdreesen:
PHPStan seems to fail, too...

falkenhawk:
> 
> 
> PHPStan seems to fail, too...

it's been fixed in #697 too.

falkenhawk:
> 
> 
> PHPStan seems to fail, too...

@jdreesen you're right, phpstan had a fix for that in https://github.com/phpstan/phpstan/pull/2275 and released in https://github.com/phpstan/phpstan/releases/tag/0.11.13, whereas php-di's dep is pinned to "^0.9.2"

falkenhawk:
```
1) DI\Test\UnitTest\Definition\InstanceDefinitionTest::should_contain_an_instance
Function ReflectionType::__toString() is deprecated

(...)/PHP-DI/vendor/phpunit/phpunit-mock-objects/src/Generator.php:919
(...)/PHP-DI/vendor/phpunit/phpunit-mock-objects/src/Generator.php:737
(...)/PHP-DI/vendor/phpunit/phpunit-mock-objects/src/Generator.php:482
(...)/PHP-DI/vendor/phpunit/phpunit-mock-objects/src/Generator.php:183
(...)/PHP-DI/vendor/phpunit/phpunit-mock-objects/src/MockBuilder.php:118
(...)/PHP-DI/vendor/phpunit/phpunit/src/Framework/TestCase.php:1497
(...)/PHP-DI/vendor/mnapoli/phpunit-easymock/src/EasyMock.php:93
(...)/PHP-DI/vendor/mnapoli/phpunit-easymock/src/EasyMock.php:33
(...)/PHP-DI/tests/UnitTest/Definition/InstanceDefinitionTest.php:26
```

The deprecation error indeed comes from phpunit, and to be more specific, from a `phpunit/phpunit-mock-objects` (which is marked as abandoned, btw). That package is required both by phpunit 6.5 and mnapoli/phpunit-easymock. @mnapoli Do you see a chance to update the test suite to newer phpunit version? But phpunit 7.x requires PHP 7.1+ and phpunit 8.x requires PHP 7.3+, so it might be complicated to have php 7.0 still supported by php-di...

mnapoli:
@falkenhawk oh wow, I didn't realized we still supported PHP 7.0.

We can jump to 7.2 without problem (the current releases are completely stable, and it's not like we are going to radically change PHP-DI and introduce major features, so people on 7.1 can still stay on the current version).

jdreesen:
I tried to upgrade `easymock` to `PHPUnit 8`: https://github.com/mnapoli/phpunit-easymock/pull/11

jdreesen:
I changed `PHP` version to `^7.2` and upgraded `PHPUnit` to `^8.5`. The tests are green now.

jdreesen:
`PHPStan` still fails on `PHP 7.4`, but an upgrade to `^0.11.13` raises 19 errors. There are some new rules, probably.

As I'm not very experienced with `PHPStan`, I just skipped its execution on `PHP 7.4` for now. Is there any advantage in running it on multiple PHP versions? If not, we can probably execute it in one PHP version (where we know it's compatible) only.

mnapoli:
Wow the work you did is huge! Thank you!

Let's keep PHPStan as it is for now, this can be done separately (later).

I'm merging and tagging a new version!

mnapoli:
Wait, actually there is no need to tag anything right now (there was nothing fixed in the code itself, it was only the tests).

I just need to remember to document that the required PHP version is up in the next release.

PR #700, opened 2019-12-14
Upgrade PHPStan to v0.12

jdreesen:
Using PHPStan's new `basline` feature, we're able to upgrade to `v0.12` right now and fix the 14 new errors later, as explained in the [first use case](https://medium.com/@ondrejmirtes/phpstans-baseline-feature-lets-you-hold-new-code-to-a-higher-standard-e77d815a5dff#fb25) of the [introduction article](https://medium.com/@ondrejmirtes/phpstans-baseline-feature-lets-you-hold-new-code-to-a-higher-standard-e77d815a5dff) of the baseline feature.

I also decreased the level from `6` to `5`, because `v0.12` introduced a [new level](https://medium.com/@ondrejmirtes/phpstan-0-12-released-f1a88036535d#1d21) that _checks for missing typehints_ which pushed previous level 6 to be level 7 now.
This new level `6` raises ~150 errors, so I think we should skip it for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ before_script:
 script:
   - if [[ $(phpenv version-name) == '7.2' ]]; then vendor/bin/phpunit --coverage-clover clover.xml ; fi
   - if [[ $(phpenv version-name) != '7.2' ]]; then vendor/bin/phpunit ; fi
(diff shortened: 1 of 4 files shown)


mnapoli:
Awesome thanks! Using the baseline for now is a good compromise.

PR #708, opened 2020-03-18
Add autowiring-support for PHP7.4 typed properties

### LINKED ISSUE #707: Support PHP7.4 typed properties in combination with `@var`
There does not seem to be support for this currently. I did a quick test - and this works in theory in `AnnotationBadedAutowiring::readProperty()`:

```php
// @Inject("name") or look for @var content
$entryName = $annotation->getName();

if ($entryName === null) {
    $entryName = $property->getType()
        ? $property->getType()->getName()
        : $this->getPhpDocReader()->getPropertyClass($property);
}
```

instead of

```php
// @Inject("name") or look for @var content
$entryName = $annotation->getName() ?: $this->getPhpDocReader()->getPropertyClass($property);
```

It obviously needs a check for versions < PHP7.4 and tests. But is this something you'd accept a PR for?

bzikarsky:
The current implementation prioritizes `@var` and the `name` property of the `@Inject` annotation over the type of the property. This allows to still inject custom-named entries into the properties even in the presence of types.

Fixes #707

### DIFF
--- a/src/Definition/Source/AnnotationBasedAutowiring.php
+++ b/src/Definition/Source/AnnotationBasedAutowiring.php
@@ -18,6 +18,7 @@ use InvalidArgumentException;
 use PhpDocReader\PhpDocReader;
 use ReflectionClass;
 use ReflectionMethod;
+use ReflectionNamedType;
 use ReflectionParameter;
 use ReflectionProperty;
 use UnexpectedValueException;
@@ -125,8 +126,17 @@ class AnnotationBasedAutowiring implements DefinitionSource, Autowiring
             return;
         }
 
-        // @Inject("name") or look for @var content
+        // Try to @Inject("name") or look for @var content
         $entryName = $annotation->getName() ?: $this->getPhpDocReader()->getPropertyClass($property);
+
+        // Try using PHP7.4 typed properties
+        if (\PHP_VERSION_ID > 70400
+            && $entryName === null
+            && $property->getType() instanceof ReflectionNamedType
+            && (class_exists($property->getType()->getName()) || interface_exists($property->getType()->getName()))
+        ) {
+            $entryName = $property->getType()->getName();
+        }
 
         if ($entryName === null) {
             throw new InvalidAnnotation(sprintf(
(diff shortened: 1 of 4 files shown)


bzikarsky:
I get a bunch of formatting errors on source files I did not touch, is this intentional?

bzikarsky:
It looks like, phpstan needs to be pushed to PHP7.4 only - I am not sure if you want me to do this in this branch? If you update master, I'll happily rebase.

mnapoli:
That pull request is awesome! Thank you for covering all cases!

I will try to sort out master, hang on ^^

mnapoli:
👍  I updated master, you can rebase/merge it in your branch.

bzikarsky:
phpstan pointed to me to two issues, I now fixed (and squashed):

1)  `getName()` only exists for `ReflectionNamedType`

2) I didn't handle scalar-typed properties "correctly" (the same way `@var` does)
    This turned out to be a bit more complex and is not 100% fixed. The separate annotation-reader project has an internal list of ignored types which I couldn't access but also didn't want to replicate. So I decided that I ignore the type on the typed property if I cannot locate a class or interface with the name. It's probably close enough to the desired behaviour. WDYT?

src/Definition/Source/AnnotationBasedAutowiring.php
```diff
@@ -125,9 +126,18 @@ private function readProperty(ReflectionProperty $property, ObjectDefinition $de
...
         }
 
-        // @Inject("name") or look for @var content
+        // Try to @Inject("name") or look for @var content
         $entryName = $annotation->getName() ?: $this->getPhpDocReader()->getPropertyClass($property);
 
+        // Try using PHP7.4 typed properties
+        if (\PHP_VERSION_ID > 70400
+            && $entryName === null
+            &&  $property->getType() instanceof ReflectionNamedType
```
mnapoli:
```suggestion
            && $property->getType() instanceof ReflectionNamedType
```

mnapoli:
Yes that looks good to me.

There is one last thing to fix regarding code formatting, I opened a code suggestion.

bzikarsky:
Thanks, missed this. PR is updated with the squashed change.

bzikarsky:
Do we have an ETA on this being merged (and tagged incl. the other PR)? :)

bzikarsky:
Ping :) -- I have a bunch of developers eagerly waiting for this feature and I'd rather not switch our projects to our forked repo.

mnapoli:
Sorry, I disable email notifications because I get so many of them every day, it creates a lot of anxiety :)

Let's give that a go!

bzikarsky:
Awesome, thanks. I'll give it a go in prod today. :)

PR #709, opened 2020-03-20
Make `injectOn` use the definition cache

bzikarsky:
Before this change `injectOn` was not using the Container internal definition cache and insetad solely built on the optional SourceCache mechanism. This is costly for setups with a long-running processes with APCu as every `injectOn` call will lead to a full annotation and autowiring run.

This change makes `injectOn` use the definition cache for all objects of non-anonymous classes. Anonymmous objects are ignored to make sure the cache does not get overloaded with many unusable definitions. It also includes a basic test which tests for minimum compatibility of `injectOn`
with anonymous classes.

This PR is a simpler alternative to #647.

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -240,7 +240,14 @@ class Container implements ContainerInterface, FactoryInterface, InvokerInterfac
             return $instance;
         }
 
-        $objectDefinition = $this->definitionSource->getDefinition(get_class($instance));
+        $className = get_class($instance);
+
+        // If the class is anonymous don't cache its definition
+        // Checking for anonymous classes is cleaner via Reflection but also less slower
+        $objectDefinition = false !== strpos($className, '@anonymous')
+            ? $this->definitionSource->getDefinition($className)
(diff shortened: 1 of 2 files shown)


src/Container.php
```diff
@@ -240,7 +240,14 @@ public function injectOn($instance)
             return $instance;
         }
 
-        $objectDefinition = $this->definitionSource->getDefinition(get_class($instance));
+        $className = get_class($instance);
+
+        // If the class is anonymous don't cache its definition
+        // Checking for anonymous classes is cleaner via Reflection but also less slower
+        $objectDefinition = false !== strpos($className, 'class@anonymous')
```
jdreesen: Note that generated names for anonymous classes change in PHP 8: https://github.com/php/php-src/pull/5153

So, to be future-proof, we should check for `@anonymous` here.

bzikarsky: True, changed. :+1:

mnapoli:
Thank you!

Issue #628, opened 2018-09-18
factory: optional parameter becomes factorydefinition


webmaster777:
While writing some code for a symfony Lock object, I was trying to configure php-di to use the factory provided for this kind of thing.

simplified:
~~~php
use Symfony\Component\Lock\StoreInterface;
use Symfony\Component\Lock\Store\FlockStore;
use Symfony\Component\Lock\LockInterface;
use Symfony\Component\Lock\Factory;

use function DI\get;
use function DI\factory;
use DI\ContainerBuilder;

$config = [
    StoreInterface::class => get(FlockStore::class),
    LockInterface::class => factory([Factory::class,"createLock"])
        ->parameter("resource", "some-lock"),
];

$containerBuilder = new ContainerBuilder();
$containerBuilder->addDefinitions($config);
$containerBuilder->useAnnotations(true);
$container = $containerBuilder->build();

$myLock = $container->get(LockInterface::class);
~~~

The definition of `Factory::createLock()` looks like this:

~~~php
    /**
     * Creates a lock for the given resource.
     *
     * @param string $resource    The resource to lock
     * @param float  $ttl         Maximum expected lock duration in seconds
     * @param bool   $autoRelease Whether to automatically release the lock or not when the lock instance is destroyed
     *
     * @return Lock
     */
    public function createLock($resource, $ttl = 300.0, $autoRelease = true)
    {
        $lock = new Lock(new Key($resource), $this->store, $ttl, $autoRelease);
        $lock->setLogger($this->logger);

        return $lock;
    }
~~~

As you can see,  only `$resource` is a required var.

However, running this will create the following Error (php7.2 / php-di 6.0.4)
~~~
PHP Fatal error:  Uncaught TypeError: Argument 3 passed to Symfony\Component\Lock\Lock::__construct() must be of the type float or null, object given, called in .../vendor/symfony/lock/Factory.php on line 47 and defined in .../vendor/symfony/lock/Lock.php:44
Stack trace:
#0 .../vendor/symfony/lock/Factory.php(47): Symfony\Component\Lock\Lock->__construct(Object(Symfony\Component\Lock\Key), Object(Symfony\Component\Lock\Store\FlockStore),Object(DI\Definition\FactoryDefinition), true)
#1 [internal function]: Symfony\Component\Lock\Factory->createLock('some-lock', Object(DI\Definition\FactoryDefinition), true)
[...]

PR #714, opened 2020-05-03
Fix issue with overriding default value of typed parameters when type is known by the container.

jworreth:
This pull request will add some more unit test for ParameterResolver class and it will fix an issue with typed parameters when the container knows the specific type which is mentioned in the function parameter. In this case, ParameterResolver will fetch the type from the container instead of setting its value to the default one.

When the type is unknown by the container, it will be forced to its default value.


mnapoli:
Closing per https://github.com/PHP-DI/PHP-DI/issues/628#issuecomment-633238067

Issue #706, opened 2020-02-26
Factories definitions using arrow functions ClosureAnalysisException


ManuelMolina97:
Hello,

We have been using PHP-DI for around a month in development environment, and everything has been good so far. We tried to save in cache our definitions:

```php

$container->addDefinitions("{$rootDir}/config/{$_SERVER['PHP_ENV']}.php");
$container->useAutowiring(true);
$container->useAnnotations(false);

$container->addDefinitions([
    \Config\DatabaseConfig::class => fn(ContainerInterface $container) => new \Config\DatabaseConfig($container->get('databaseConfig')),
...
]);

    $container->enableCompilation(__DIR__ . '/cache');
    $container->writeProxiesToFile(true, __DIR__ . '/cache/proxies');
``` 

And when `writeProxiesToFile` was being executed, we got the following error:

```
2020/02/26 10:02:31 [error] 1100#1100: *3 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Uncaught SuperClosure\Exception\ClosureAnalysisException: The closure was not found within the abstract syntax tree. in /var/www/wallet/vendor/jeremeamia/superclosure/src/Analyzer/AstAnalyzer.php:72
Stack trace:
#0 /var/www/pracmanduca/vendor/jeremeamia/superclosure/src/Analyzer/AstAnalyzer.php(30): SuperClosure\Analyzer\AstAnalyzer->locateClosure()
#1 /var/www/pracmanduca/vendor/jeremeamia/superclosure/src/Analyzer/ClosureAnalyzer.php(29): SuperClosure\Analyzer\AstAnalyzer->determineCode()
#2 /var/www/pracmanduca/vendor/php-di/php-di/src/Compiler/Compiler.php(373): SuperClosure\Analyzer\ClosureAnalyzer->analyze()
#3 /var/www/pracmanduca/vendor/php-di/php-di/src/Compiler/Compiler.php(322): DI\Compiler\Compiler->compileClosure()
#4 /var/www/pracmanduca/vendor/php-di/php-di/src/Compiler/Compiler.php(276): DI\Compiler\Compiler->compileValue()
#5 /var/www/pracmanduca/vendor/php-di/php-di/src/Compiler/Compiler.php(155): DI\Compiler\Compiler->compileDefinition()
#6 /var/www/pracmanduca/vendor/php-di/php-di/
```

[...]

PR #716, opened 2020-05-17
Add support for compilation of the short closures

### LINKED ISSUE #711: Dependecy jeremeamia/superclosure is abandoned
When running `composer outdated` on my project, I'm receiving message
```
Package jeremeamia/superclosure is abandoned, you should avoid using it. Use opis/closure instead.
```
Similar message is on package [it's page](https://packagist.org/packages/jeremeamia/superclosure).

opis/closure apparently comes with functionality comparable to SuperClosure as noted in [docs](https://docs.opis.io/closure/3.x/#superclosure-support).

b1rdex:
Fixes #711, #706.

It replaces jeremeamia/superclosure with opis/closure and drops nikic/php-parser dependency.

There are two problems with this:

1. looks like there is a bug regarding the detection of scope references usage – it treats `static` variable definition as a scope reference. See https://github.com/opis/closure/issues/52 for details.

2. it doesn't throw an exception on multiple closures defined on the same line. Just silently returns the code of the first one.

If anyone has free time – please create PRs in opis/closure to fix the problems.

I haven't tested this with real container definitions, so feel free to test and report issues.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,6 @@
         "psr/container": "^1.0",
         "php-di/invoker": "^2.0",
         "php-di/phpdoc-reader": "^2.0.1",
-        "nikic/php-parser": "^2.0|^3.0|^4.0",
         "opis/closure": "^3.5"
     },
     "require-dev": {
--- a/composer.json
+++ b/composer.json
@@ -29,8 +29,8 @@
         "psr/container": "^1.0",
         "php-di/invoker": "^2.0",
         "php-di/phpdoc-reader": "^2.0.1",
-        "jeremeamia/superclosure": "^2.0",
-        "nikic/php-parser": "^2.0|^3.0|^4.0"
+        "nikic/php-parser": "^2.0|^3.0|^4.0",
+        "opis/closure": "^3.5"
     },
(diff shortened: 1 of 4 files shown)


b1rdex:
P. 1 isn't going to be fixed. Please see https://github.com/opis/closure/issues/52#issuecomment-632344526 for details.
So we have to restrict the usage of `static` variables.

mnapoli:
Thank you for following up with all that.

Let's start with the easy one: multiple closures on one line. That should never happen in container configuration, and even then, users currently have errors, so it's not a regression. I'm fine with that.

Regarding `static` variables: removing support for that would be a BC break. However, it wouldn't make sense to me that these would be actually used. But how to be certain? 🤔  Except the BC break, I am perfectly fine with not supporting these.

Tagging a major PHP-DI version for just that is not great. I asked for clarification in https://github.com/opis/closure/issues/52#issuecomment-632566611

msarca:
We had a small issue with `null` scopes but that was fixed in v3.5.3. We do no longer use the class scope optimisation so the issue with `$isUsingScope` do not exist anymore.

PR #718, opened 2020-06-13
Replace deprecated SimpleAnnotationReader

astehlik:
The `SimpleAnnotationReader` of doctrine is deprecated / removed
in future versions. See also:

https://github.com/doctrine/annotations/issues/87

It is replaced with the `AnnotationReader`.

This is a breaking change because annotations without proper
namespacing are no longer supported!

I also looked for a backwards compatible solution but I could not find one that made sense :( 

Drupal for example seems to have copied the source of the `SimpleAnnotationReader` into their own project: https://www.drupal.org/project/drupal/issues/2631202

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -34,7 +34,7 @@
     "require-dev": {
         "phpunit/phpunit": "^8.5",
         "mnapoli/phpunit-easymock": "^1.2",
-        "doctrine/annotations": "~1.2",
+        "doctrine/annotations": "^1.10",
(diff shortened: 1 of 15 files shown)


mnapoli:
For PHP-DI 7 I want to replace annotations entirely with attributes: #812 

That should solve this issue 👍

astehlik:
:+1: Sounds good, closing this PR

PR #721, opened 2020-07-17
Moved inline namespace calls to use

iruzevic:
We are using you lib WordPress project and the issue arises when we have multiple composer packages (plugin and theme), we are using https://github.com/TypistTech/imposter extension for adding vendor prefix based on the project.

I have seen that on few places in the project you have inline class calls using full namespace name. This presents and issue for me. So I created this PR to move those calls to top of the class and call the classes using short name.

All tests are passing


mnapoli:
👍  if it helps you, I don't see any harm in this change.

iruzevic:
Thank you very much

PR #728, opened 2020-08-23
Make Compiler save file atomically

nightlinus:
As of now compiled container can be broken when concurrently compile container.
This pr add saving compiled content to temporary file and then rename it to destination file.

### DIFF
--- a/src/Compiler/Compiler.php
+++ b/src/Compiler/Compiler.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace DI\Compiler;
 
+use function chmod;
 use DI\Definition\ArrayDefinition;
 use DI\Definition\DecoratorDefinition;
 use DI\Definition\Definition;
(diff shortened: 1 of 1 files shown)


holtkamp:
Regarding https://github.com/PHP-DI/PHP-DI/pull/728#issuecomment-678794475, I would say it _is_ worth the extra package/dependency because:

1. this kind of functionality might introduce "strange" behavior, such as: https://github.com/webimpress/safe-writer/issues/2
2. both [`ocramius/proxy-manager`](https://packagist.org/packages/ocramius/proxy-manager) and [`php-di/php-di`](https://packagist.org/packages/php-di/php-di) are used a lot, so sharing the this dependency only makes it more reliable, right and tested all over the world in different setups.
3. the package itself does not have any additional dependencies: https://github.com/webimpress/safe-writer/blob/e185e55135c240dc5f490cae8ea3ae8051959db4/composer.json#L12-L14
4. and maybe most important: it is well-tested: https://github.com/webimpress/safe-writer/tree/master/test

Just my two cents ;)

PR #738, opened 2020-09-28
Inject via PHP 8 attributes

mnapoli:
PHP-DI 7 will support PHP 8 attributes as a replacement for Doctrine annotations.

I think it still makes sense to support Doctrine annotations if we can to ease upgrading, so I haven't removed it. They are however deprecated.

Example:

```php
use DI\Attribute\Inject;

class UserController
{
    #[Inject]
    private FormFactoryInterface $formFactory;

    public function createForm($type, $data, $options)
    {
        // $this->formFactory->...
    }
}
```

The attribute behaves like the annotation. Note that the class is different (`DI\Attribute\Inject` instead of `DI\Annotation\Inject`).

The "best practices" recommendations do not change: using attributes is recommended in places like controllers and other glue code, and discouraged in domain code.

### DIFF
--- a/couscous.yml
+++ b/couscous.yml
@@ -46,6 +46,9 @@ menu:
                 php-definitions:
                     text: PHP definitions
                     url: doc/php-definitions.html
+                attributes:
+                    text: PHP 8 attributes
+                    url: doc/attributes.html
                 annotations:
                     text: Annotations
                     url: doc/annotations.html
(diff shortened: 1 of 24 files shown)


mnapoli:
The build is currently failing because PHPStan doesn't fully support PHP 8 yet, and phpcs doesn't support attributes too.

Good to merge in branch v1.0.

PR #742, opened 2020-10-28
Update .gitattributes

KarelWintersky:
Remove useless files for installed package

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,9 +1,21 @@
-# .gitattributes
-.github/ export-ignore
-tests/ export-ignore
-website/ export-ignore
-doc/ export-ignore
-news/ export-ignore
-
 # Auto detect text files and perform LF normalization
 * text=auto
+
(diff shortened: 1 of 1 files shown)


.gitattributes
```diff
@@ -1,9 +1,21 @@
...
+.gitstats.yml     export-ignore
+.php_cs           export-ignore
+.phpstorm.meta.php export-ignore
+404.md            export-ignore
+CONTRIBUTING.md   export-ignore
+couscous.yml      export-ignore
+*.neon            export-ignore
+phpunit.xml.dist  export-ignore
+SECURITY.md       export-ignore
+
```
mnapoli, line 21: Could you keep the original location so that the diff is valid, I see no reason to move the lines?

KarelWintersky: May be. Check pls.

mnapoli:
Thanks!

PR #750, opened 2020-12-10
Add ContainerInterface

dflourusso:
Useful to obtain the autocomplete of the "set" method that does not exists in PSR ContainerInterface

**From:**
image

**To:**
image


mnapoli:
I'm sorry but that doesn't make any sense, I'm not sure which problem you are trying to solve here.

That interface would only be implemented by PHP-DI, you are not decoupling anything.

SvenRtbg:
My suggestion would be: If you want to call `set()` on a container object, don't typehint the PSR `ContainerInterface`, just typehint the container class you are using. You already have a dependency on that nonstandard method anyways.

dflourusso:
In some cases I need to pass the container as dependency injection, and in
my test files I want to pass a fake container. It would be easy to do with
an interface typehint and not the container for itself.

Em sex., 11 de dez. de 2020 às 13:05, SvenRtbg <notifications@github.com>
escreveu:

PR #755, opened 2020-12-17
Update docs (php-definitions: static methods)

### LINKED ISSUE #752: Question about php definitions, factory and static method
Hi!

In the docs, [php definitions -> factories](https://php-di.org/doc/php-definitions.html#factories), there is the following mention:

```
factory([FooFactory::class, 'build']): if build() is a static method then the object will not be created: FooFactory::build() will be called statically (as one would expect)
```

~I don't understand why this wouldn't work as:~ I don't understand why new object would not be created if the following works in vanilla php:

```php
class Test {

  public static function factory(): self
  {
    return new Test();
  }

}

var_dump(Test::factory());
```

returns `object(Test)[1]`.

Did you mean something else here?

alexkuc:
As discussed earlier with @mnapoli and @SvenRtbg, this updates description of php definitions for static factory (fix #752)

### DIFF
--- a/doc/php-definitions.md
+++ b/doc/php-definitions.md
@@ -213,7 +213,7 @@ return [
 
 Please note:
 
-- `factory([FooFactory::class, 'build'])`: if `build()` is a **static** method then the object will not be created: `FooFactory::build()` will be called statically (as one would expect)
+- `factory([FooFactory::class, 'build'])`: if `build()` is a **static** method then `FooFactory::build()` will be called statically (the static method is responsible for returning an object or returning a value)
 - you can set any container entry name in the array, e.g. `DI\factory(['foo_bar_baz', 'build'])` (or alternatively: `DI\factory('foo_bar_baz::build')`), allowing you to configure `foo_bar_baz` and its dependencies like any other object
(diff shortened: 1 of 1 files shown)


doc/php-definitions.md
```diff
@@ -213,7 +213,7 @@ return [
 
 Please note:
 
-- `factory([FooFactory::class, 'build'])`: if `build()` is a **static** method then the object will not be created: `FooFactory::build()` will be called statically (as one would expect)
+- `factory([FooFactory::class, 'build']):` if `build()` is a **static** method then `FooFactory::build()` will be called statically (the static method is responsible for returning an object or returning a value)
```
mnapoli, line 216: Let's revert the type at the beginning:

```suggestion
- `factory([FooFactory::class, 'build'])`: if `build()` is a **static** method then `FooFactory::build()` will be called statically (the static method is responsible for returning an object or returning a value)
```

alexkuc: I have merged your changes with squash (to keep only 1 commit). Let me know if this looks okay now.

mnapoli:
Thank you!

Issue #754, opened 2020-12-16
How do I provide a variableReader callable for the EnvironmentVariableResolver


angelxmoreno:
Looking at the [EnvironmentVariableResolver class](https://github.com/PHP-DI/PHP-DI/blob/0ca3e76b3ce6f61f1cc2b8c881bf18f14a49e440/src/Definition/Resolver/EnvironmentVariableResolver.php#L16), I notice it has a default `variableReader` of `getenv`. Hod Do I change this?

mnapoli:
I don't know if that's possible, what's you objective?

angelxmoreno:
my objective is to use a thread safe method to get my env vars. `getenv` is not thread-safe and thus, vlucas/phpdotenv deprecated `putenv` : https://github.com/vlucas/phpdotenv#putenv-and-getenv

SvenRtbg:
I'd assume reading something is always thread safe, at least if it isn't changed. It is writing to something that makes things interesting.

Are you intending to change these environment variables? How?

mnapoli:
Then maybe the best course of action would be to change PHP-DI to use a thread safe method?

SvenRtbg:
Well, the discussion about that feature seems to be summarized in this statement: https://github.com/vlucas/phpdotenv/issues/446#issuecomment-639016097

So to me it seems that thread-safety only applies to certain ways of running PHP (well, obviously it has to run in threads, so PHP-FPM is mentioned), then additionally the environment variables have to be different in order to "leak" between threads. If you always look at the same environment variables and also do not attempt to change it from a php script, then variable content may still leak, but you probably wouldn't notice. However, PHP-DI must not attempt to assume what the users are running their scripts on.

The alternative choice should be to access `$_SERVER`. Which probably is a good idea anyways, but I would think that this access method is already possible within PHP-DI: Just set a value, populate it with `'database.host' => $_SERVER['YOUR_DB_HOST_ENV_VAR']` and use that value where needed, instead of using the `\DI\Env` helper.

mnapoli:
@SvenRtbg I'm not sure I understand: why shouldn't PHP-DI use `$_SERVER` instead of `getenv()` if `$_SERVER` is better?

SvenRtbg:
Like you, I am still trying to understand the use case. However I was also suggesting an alternative to work with env vars that already works. Not intending to stop progress. :)

mnapoli:
No worries, just wanted to make sure I'm not missing anything.

[...]

PR #757, opened 2020-12-29
Feature/754 replace getenv

angelxmoreno:
Based on discussions in #754, replaced the default method for accessing environment variables.

### DIFF
--- a/src/Definition/Resolver/EnvironmentVariableResolver.php
+++ b/src/Definition/Resolver/EnvironmentVariableResolver.php
@@ -31,11 +31,6 @@ class EnvironmentVariableResolver implements DefinitionResolver
         $this->variableReader = $variableReader ?? [$this, 'getEnvVariable'];
     }
 
-    protected function getEnvVariable(string $variableName)
-    {
-        return isset($_SERVER[$variableName]) ? $_SERVER[$variableName]: false;
-    }
-
     /**
      * Resolve an environment variable definition to a value.
      *
@@ -66,8 +61,13 @@ class EnvironmentVariableResolver implements DefinitionResolver
         return $value;
     }
 
-    public function isResolvable(Definition $definition, array $parameters = []) : bool
+    public function isResolvable(Definition $definition, array $parameters = []): bool
     {
         return true;
+    }
+
+    protected function getEnvVariable(string $variableName)
+    {
+        return isset($_SERVER[$variableName]) ? $_SERVER[$variableName] : false;
     }
 }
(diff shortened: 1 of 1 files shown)


src/Definition/Resolver/EnvironmentVariableResolver.php
```diff
@@ -65,4 +65,9 @@ public function isResolvable(Definition $definition, array $parameters = []) : b
     {
         return true;
     }
+
+    protected function getEnvVariable(string $variableName)
+    {
+        return $_SERVER[$variableName] ?? false;
```
mnapoli, line 71: After several tests unfortunately it seems that `$_SERVER` is not always populated by PHP. For example on macOS, after installing PHP with Brew, I don't get the variables.

I'm not sure how other frameworks deal with it but I remember having this exact problem with Symfony in the past.

MrPropre: @mnapoli You could maybe try checking the php.ini file to see if `variables_order` contains "S" for the super global `$_SERVER`.
Additionally, if `auto_globals_jit` is enabled, PHP.NET says that “the SERVER and ENV variables are created when they're first used (Just In Time) instead of when the script starts.". So if these variables are not used in a script, they will not be initialized.

MrPropre: Also, looking at the [EnvVarProcessor](https://github.com/symfony/dependency-injection/blob/c1af27be7ca50969ee4e6cbefb3107715a2070b2/EnvVarProcessor.php#L131-L135) class of Symfony, we can see that it checks, in order, `$_ENV`, `$_SERVER`, `getenv()` and the .env file.

mnapoli: I know that this is something that I can configure, but I'm talking about users of PHP-DI.

I want this to work out of the box for everyone, not just me.

MrPropre: @mnapoli So instead, as a workaround:
```suggestion
        return $_ENV[$variableName] ?? $_SERVER[$variableName] ?? getenv($variableName);
```
It will suit most users.

mnapoli: Yes that sounds like a really good suggestion, and Symfony is doing exactly that!

angelxmoreno:
Applied the suggestions! Thanks so much guys for the review

mnapoli (approved):
Thanks, looking good!

The code sniffer build is failing though, if you can fix that I'll merge it!

src/Definition/Resolver/EnvironmentVariableResolver.php
```diff
@@ -65,4 +65,14 @@ public function isResolvable(Definition $definition, array $parameters = []) : b
...
     }
+
+    protected function getEnvVariable(string $variableName)
+    {
+        if (isset($_ENV[$variableName])) {
+            return $_ENV[$variableName];
+        } elseif (isset($_SERVER[$variableName])) {
+            return $_SERVER[$variableName];
+        }
+        return getenv($variableName);
```
MrPropre, line 76:
```suggestion

        return getenv($variableName);
```

mnapoli:
We're all good, thanks!

PR #761, opened 2021-01-19
stop using assertRegExp in favor of assertMatchesRegularExpression

partikus:
More details can be found at https://github.com/sebastianbergmann/phpunit/issues/4086

edit:
I've overlooked that `assertMatchesRegularExpression` works only with phpunit 9.x, so it cannot be merged until we switch to `phpunit/phpunit: ^9.0|^10.0`.


mnapoli:
Yep we cannot upgrade since we support PHP 7.2, gotta close this for now.

PR #764, opened 2021-01-21
adding psr container exception to the ignored errors list

partikus:
Since phpstan throws an error for a PSR container exception interface we have to add this class to the ignored list.

https://github.com/phpstan/phpstan/issues/1066
https://github.com/php-fig/container/issues/21


mnapoli:
Thank you!

PR #766, opened 2021-02-27
Support PHP 8 only in PHP-DI 7

mnapoli:
Initially, the plan was to:

- support PHP 7.2 -> 7.4 in PHP-DI 6
- support PHP 7.4 -> 8.0+ in PHP-DI 7

In the meantime, I added PHP 8 support in PHP-DI 6.

Since PHP-DI 7 is coming slower than I anticipated, I'm considering dropping support for PHP 7 entirely. That allows to add stricter types to PHP-DI, along with stronger Psalm typing, as well as reduce effort maintenance. PHP-DI 6 is still supported, and very stable, so it's not like users would be forced to upgrade to PHP-DI 7 immediately.

I'm going to leave this PR open for a little while if there are some opinions on that.


SvenRtbg:
It's not that you have an army of volunteers and can afford a regular release schedule like PHPUnit does, for example. If you can still support older maintained PHP versions in any version, I am fine with that.

mnapoli:
Yep, and it's not like I have huge plans for PHP-DI 7. What I love about this project is that it's now very mature, there isn't much to break or to add.

So PHP-DI 6 and 7 would be very similar, which means that it's OK to stay on PHP-DI 6 a bit longer if you can't upgrade to PHP 8.

Thanks for the feedback!

PR #777, opened 2021-04-30
Get env vars from $_ENV and $_SERVER in compiled container

### LINKED ISSUE #776: CompiledContainer does not look in $_ENV and $_SERVER for env variables
Hello,

When compilation is enabled on the container builder, the compiled container only use the `getenv` function to get env variables.
```php
    protected function get2()
    {
                $value = getenv('DB_HOST');
        if (false !== $value) return $value;
        if (!false) {
            throw new \DI\Definition\Exception\InvalidDefinition("The environment variable 'DB_HOST' has not been defined");
        }
        return NULL;
    }
````
I think it should also check in $_ENV and $_SERVER variables before, since the default method for accessing env variables has been changed: #757 

A simple way to do this without adding too much code would be to use `??` (null coalescing operator: `$_ENV['DB_HOST'] ?? $_SERVER['DB_HOST'] ?? getenv('DB_HOST')`) but this would not work for empty strings (`DB_HOST=""`). Maybe you should write a custom function instead.

I don't know what is the better solution, so I prefer submitting an Issue instead of a PR.

Thank you in advance!

rdarcy1:
Closes #776 

Brings the compiled container in line with the non-compiled environment variable resolver (#757). Uses thread-safe `$_ENV` and `$_SERVER` in preference to `getenv`.

I couldn't see any tests relating specifically to this part but happy to add one if needed.

### DIFF
--- a/src/Compiler/Compiler.php
+++ b/src/Compiler/Compiler.php
@@ -244,7 +244,7 @@ class Compiler
                 $isOptional = $this->compileValue($definition->isOptional());
                 $defaultValue = $this->compileValue($definition->getDefaultValue());
                 $code = <<<PHP
-        \$value = getenv($variableName);
+        \$value = \$_ENV[$variableName] ?? \$_SERVER[$variableName] ?? getenv($variableName);
         if (false !== \$value) return \$value;
         if (!$isOptional) {
             throw new \DI\Definition\Exception\InvalidDefinition("The environment variable '{$definition->getVariableName()}' has not been defined");


mnapoli:
Thank you!

shadowhand:
Is it safe to use `$_SERVER` this way, given that it can be modified by HTTP request?

rdarcy1:
I believe the `$_ENV` superglobal can be modified in the same way? I don't have strong feelings about it, but both Laravel (vlucas/phpdotenv) and Symfony (symfony/dotenv) allow the use of `$_SERVER` so it's pretty well-established.

PR #778, opened 2021-05-04
Improve DI\add description

### LINKED ISSUE #758: Overriding array values recusrively
Related to overriding array values as described [here](https://php-di.org/doc/definition-overriding.html#arrays):

This doesn't work if you're wanted to override a nested array. For example, I define default settings in one definition file...

```php
return [
    'settings' => [
        // Database settings
        'db' => [
            'host' => 'localhost',
            'user' => '',
            'pass' => '',
            'dbname' => 'test',
        ]
];
```

... and then environment specific settings in another file to override the default

```php
return [
    'settings' => DI\add([
        // Database settings
        'db' => [
            'user' => 'REALUSERNAME',
            'pass' => 'REALPASSWORD',
        ]
    ])
];
```

I'd expect to end up with 
```php
'settings' => [
   'db' => [
      'host' => 'localhost',
      'user' => 'REALUSERNAME',
      'pass' => 'REALPASSWORD',
      'dbname' => 'test',
   ]
]
```

But instead the second value just overwrites the first
```php
'settings' => [
   'db' => [
      'user' => 'REALUSERNAME',
      'pass' => 'REALPASSWORD'
   ]
]
```

I'm guessing the solution to this is modifying https://github.com/PHP-DI/PHP-DI/blob/master/src/Definition/ArrayDefinitionExtension.php#L28 to manually walk the array and merge the values of any nested `ArrayDefinition` instead of using `array_merge`. 

Or another solution would to be update the documentation to say overriding nested values isn't supported 😀. It's not difficult to switch to values such as `db.host` instead of actual nested arrays and the overriding works fine once I do this, I guess it just wasn't obvious for a newbie reading the documentation.

Thanks!

SvenRtbg:
Resolves #758

Document that DI\add does not support recursive extension of arrays.

### DIFF
--- a/doc/definition-overriding.md
+++ b/doc/definition-overriding.md
@@ -119,6 +119,18 @@ When resolved, the array will contain the 2 entries. **If you forget to use `DI\
 
 Note that you can use `DI\add()` even if the array was not declared before.
 
+Adding to arrays in the DI definition does not work recursively, i.e. having an array entry that contains an array.
+
+```php
+return [
+    'array' => [
+        'subarray' => [DI\get(Entry::class)],
+    ],
+];
+```
+
+You cannot add elements to `subarray` using `DI\add()`.
+
 ### Decorators
 
 You can use `DI\decorate()` to decorate an object:


mnapoli:
Makes sense thank you!

Issue #773, opened 2021-04-19
Broken Link: Integration - Zend Expressive


arueckauer:
On the [main page](https://php-di.org/) in the Integrations section, there is a link for Zend Expressive. The referenced document (https://php-di.org/doc/frameworks/ze.html) can not be found.

mnapoli:
Good point, that should be `zend-expressive` instead of `ze`.

If anyone wants to send a pull request that would be awesome.

PR #784, opened 2021-05-24
Broken Link: Integration - Zend Expressive #773

elie29:


couscous.yml
```diff
@@ -67,7 +67,7 @@ menu:
                 zf1:
                     text: Zend Framework 1
                     url: doc/frameworks/zf1.html
-                ze:
+                zend-expressive:
```
mnapoli, line 70: I'm not sure about changing this, the `ze` key is used in the header of https://github.com/PHP-DI/PHP-DI/blob/master/doc/frameworks/zend-expressive.md too

elie29: @mnapoli thx for the review, but we could have the same issue with symfony, the key is symfony and not symfony2. However, i will make the change for ze in the menu

mnapoli:
Thanks!

PR #786, opened 2021-06-09
Atomic creation of compilation directory

### LINKED ISSUE #785: Creating compilation directory is not atomic
Creating the compilation directory for compiled file with definitions is not atomic and fails, if multiple instances of PHP are executed in parallel.

**How to reproduce:**

1) Create a sample script:
```php
<?php

require_once __DIR__ . "/vendor/autoload.php";

$builder->addDefinitions(['foo' => 'bar']);
$builder->enableCompilation(__DIR__ . "/tmp");
$builder->build();
```

2) Simulate slow disk by adding delay between `is_dir()` and `mkdir()` in `Compiler::createCompilationDirectory` (https://github.com/PHP-DI/PHP-DI/blob/master/src/Compiler/Compiler.php#L361):
```php
if (!is_dir($directory) && (sleep(1) || true) && !@mkdir($directory, 0777, true)) {
//                         ^^^^^^^^^^^^^^^^^^
```

3) Run the sample script multiple times in parallel:
```shell
rm -rf tmp/ && php script.php & php script.php && wait
```

Exception is thrown `Uncaught InvalidArgumentException: Compilation directory does not exist and cannot be created`.

If this is accepted as a bug, I am willing to create a patch for a fix.

martin-heralecky:
closes #785

Added additional check for whether the directory exists. This way, if the `mkdir()` fails, but the directory does exists afterwards (was presumably created by another script running at the same time), no exception is thrown.

### DIFF
--- a/src/Compiler/Compiler.php
+++ b/src/Compiler/Compiler.php
@@ -358,7 +358,7 @@ PHP;
 
     private function createCompilationDirectory(string $directory)
     {
-        if (!is_dir($directory) && !@mkdir($directory, 0777, true)) {
+        if (!is_dir($directory) && !@mkdir($directory, 0777, true) && !is_dir($directory)) {
             throw new InvalidArgumentException(sprintf('Compilation directory does not exist and cannot be created: %s.', $directory));
         }
         if (!is_writable($directory)) {


martin-heralecky:
I'd greatly appreciate, if this could be patched to 6.* as well.

mnapoli:
Makes sense, thank you!

I'll merge and release master, which is 6.x

PR #790, opened 2021-08-08
Mention that DI\get() can also be used to map interfaces

drealecs:
I've been using `DI\get` to map interface to implementation but documentation only mention `DI\create` and `DI\autowire` for this purpose.
The advantage of using `DI\get` is that when requesting either the class or the interface mapped to it, only one object will be created by the container and returned.
When autowiring is enabled, that just means using `get` instead of `autowire` if there are no other configuration for the entry.

This just adds to the documentation that `get` can be used for mapping interfaces to implementation using roughly the same wording as for `create` and `autowire`.
Please let me know if the changes are capturing this well enough.

### DIFF
--- a/doc/php-definitions.md
+++ b/doc/php-definitions.md
@@ -367,6 +367,15 @@ return [
 ];
 ```
 
+You can use aliases to map interfaces to implementation obtaining the same object when requesting/injecting the class or the interface
+
+```php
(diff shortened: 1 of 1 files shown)


mnapoli:
Thank you!

I do use that too, that makes sense.

PR #792, opened 2021-09-01
Add @template return types for better autocomplete/static analysis

rdarcy1:
PhpStorm now supports Psalm-style [class string](https://psalm.dev/docs/annotating_code/type_syntax/scalar_types/#class-string-interface-string) and [template](https://psalm.dev/docs/annotating_code/templated_annotations/) annotations ([announcement](https://blog.jetbrains.com/phpstorm/2021/07/phpstorm-2021-2-release/)). This works by default without needing to install or run Psalm on the project separately.

By adding template annotations to `get`, `make` and `injectOn` methods, we can get autocomplete without any inline docblocks or extra metadata files. 

I think by keeping the existing type annotations and just unioning the new template types, it should work fine with IDEs that don't support this syntax (note I haven't tested on any other IDEs).

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -117,11 +117,12 @@ class Container implements ContainerInterface, FactoryInterface, InvokerInterfac
     /**
      * Returns an entry of the container by its name.
      *
-     * @param string $name Entry name or a class name.
+     * @template T
+     * @param string|class-string<T> $name Entry name or a class name.
      *
      * @throws DependencyException Error while resolving the entry.
      * @throws NotFoundException No entry found for the given name.
-     * @return mixed
+     * @return mixed|T
      */
     public function get($name)
     {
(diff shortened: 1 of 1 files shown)


mnapoli:
Thank you! I have one question: does that PR still supports non-class-string parameters/returned values?

Because yes, 90% of services are usually "class name -> class instance" mapping. But we don't want to break anything for services and values that are not indexed by a class name.

Here's an example to illustrate:

```php
$container->set('foo', 'bar');

$abc = $container->get('foo');
```

Will static analysis report any issue?

rdarcy1:
That's a good point – in that case the return type union with `mixed` covers it and the IDE just doesn't provide any completion.

Quick demo here, no errors on string/array and autocomplete on the object:

Screenshot 2021-09-02 at 10 39 32

mnapoli:
Looks good, thank you for trying it out!

Let's give this a try. If there's ever an issue that we did not anticipate, it will not be the kind of issue that will break production suddenly, so I'm fine with releasing now and iterate if needed.

Thanks for the PR!

PR #793, opened 2021-09-14
Makes serializable closure work with PHP 8.1

nunomaduro:
This pull request makes PHP DI support PHP 8.1 regarding serializable closures. Waiting for a stable version of Laravel Serializable Closures.

Issue #727, opened 2020-08-05
ReflectionParameter::getClass is deprecated in PHP 8


timfennis:
I'm testing out PHP-DI on PHP 8 and noticed that it's throwing some deprecation warnings. If this issue is not wanted feel free to delete it.

```
Deprecated: Method ReflectionParameter::getClass() is deprecated in /var/www/vendor/php-di/php-di/src/Definition/Source/ReflectionBasedAutowiring.php on line 65
```

and

```
Deprecated: Method ReflectionParameter::getClass() is deprecated in /var/www/vendor/php-di/php-di/src/Invoker/FactoryParameterResolver.php on line 45
```

timfennis:
In ReflectionBaseduaowiring.php chaning getparamtersDefinition to something like

```
    private function getParametersDefinition(\ReflectionFunctionAbstract $constructor) : array
    {
        $parameters = [];

        foreach ($constructor->getParameters() as $index => $parameter) {
            // Skip optional parameters
            if ($parameter->isOptional()) {
                continue;
            }

            $type = $parameter->getType();
            $class = null !== $type && !$type->isBuiltin() ? $type->getName() : null;

            if ($class) {
                $parameters[$index] = new Reference($class);
            }
        }

        return $parameters;
    }
```

seems to work. (Solution based on pull-requests seen elsewhere).

For FactoryParamterResolver.php the following solution appears to work:

```
    public function getParameters(
        ReflectionFunctionAbstract $reflection,
        array $providedParameters,
        array $resolvedParameters
    ) : array {
        $parameters = $reflection->getParameters();

        // Skip parameters already resolved
        if (! empty($resolvedParameters)) {
            $parameters = array_diff_key($parameters, $resolvedParameters);
        }

        foreach ($parameters as $index => $parameter) {
            $type = $parameter->getType();
            $class = null !== $type && !$type->isBuiltin() ? $type->getName() : null;

            if (!$class) {
                continue;
            }

            if ($class === 'Psr\Container\ContainerInterface') {
                $resolvedParameters[$index] = $this->container;
            } elseif ($class === 'DI\Factory\RequestedEntry') {
                // By convention the second parameter is the definition
                $resolvedParameters[$index] = $providedParameters[1];
            } elseif ($this->container->has($class)) {
                $resolvedParameters[$index] = $this->container->get($parameterClass->name);
            }
        }

        return $resolvedParameters;
    }
```

[...]

PR #794, opened 2021-09-15
Support PHP 8.1 by moving from opis/closure to laravel/serializable-closure

mnapoli:
This is a continuation of #793 with fixed tests and complete removal of opis/closure.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,6 @@
         "psr/container": "^1.0",
         "php-di/invoker": "^2.0",
         "php-di/phpdoc-reader": "^2.0.1",
-        "opis/closure": "^3.5.5",
         "laravel/serializable-closure": "^1.0"
     },
     "require-dev": {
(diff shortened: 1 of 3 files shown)


mnapoli:
🤦  I completely forgot we still supported PHP 7.2. Let me try that again…

hemberger:
Hello! Since PHP 7.2 reached EOL over a year ago (2020-11-30), is there a reason to continue supporting it in future PHP-DI versions? Personally, I would be happy to see the minimum req bumped to 7.3 so that the deprecation issue in PHP 8.1 can be addressed. But since I am not involved in the PHP-DI development in any way, I don't know if this (or some other solution) is already on the horizon. Thank you for your time!

mnapoli:
Matomo still supports PHP 7.2: https://github.com/PHP-DI/PHP-DI/issues/727#issuecomment-703939607

Ideally I'd like to find a solution to still support it (if possible)

PR #795, opened 2021-10-01
Hint on variadic parameters for `addDefinitions`

tomkyle:
Basic example in docs shows one definition file passed to _addDefinitions_, so one might be tempted to call _addDefinitions_ repeatedly. By using variadic parameters with `...`, adding a bunch of definition files becomes possible.

### DIFF
--- a/doc/php-definitions.md
+++ b/doc/php-definitions.md
@@ -28,6 +28,17 @@ return [
 $containerBuilder->addDefinitions('config.php');
 ```
 
+The _addDefinitions_ method is variadic, so it accepts multiple definition files:
+
+```php
+$containerBuilder->addDefinitions( ...[
+    'config.php', 
+    'actions.php'
+]);
+```
+
+
+
 ## A word about lazy loading
 
 PHP-DI loads the definitions you have written and uses them like instructions on how to create objects.


mnapoli:
Thanks.

Honestly that would work too, and be simpler:

```php
$containerBuilder->addDefinitions('config.php', 'actions.php');
```

but TBH I don't think it's worth adding to the documentation. It's something that can be discovered via the IDE, and it's not critical (compared to using 2 method calls). I'd rather keep the documentation as light as possible.

tomkyle:
Thanks for your thoughts and fast reaction; agree with you that simple docs are better docs. Not sure about your IDE argument, when I look at Sublime Text and other bare-bone users. In that sense, mentioning the variadic feature would be helpful. – BTW my example was in that way misleading, as it's not about adding _two_ such files (in fact I forgot to add `...`). The who split their definitions in such files, easily end up with 10 to 20 files for each “reuseable source code topic”.

mnapoli:
> Not sure about your IDE argument, when I look at Sublime Text and other bare-bone users.

phpdoc is part of the API and documentation, if some users don't have autocompletion available I'm afraid there's not much we can do to help them. It's like if they didn't have a web browser to read the docs 😄  Things will naturally be harder for them.

PR #808, opened 2022-03-09
Switch closure serializer for PHP 8.1 support

### LINKED ISSUE #791: php 8.1 support with opis/closure:4.x-dev
php-di now can't support closure serialize with php 8.1, need depend on opis/closure:4.x-dev.

shadowhand:
Fixes #791

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,17 +25,18 @@
         "phpstan": "phpstan analyse -l 5 -c phpstan.neon src"
     },
     "require": {
-        "php": ">=7.2.0",
+        "php": ">=7.3.0",
         "psr/container": "^1.0",
         "php-di/invoker": "^2.0",
         "php-di/phpdoc-reader": "^2.0.1",
-        "opis/closure": "^3.5.5"
+        "opis/closure": "^3.5.5",
+        "laravel/serializable-closure": "^1.0"
     },
     "require-dev": {
         "phpunit/phpunit": "^8.5|^9.0",
         "mnapoli/phpunit-easymock": "^1.2",
         "doctrine/annotations": "~1.2",
-        "ocramius/proxy-manager": "^2.0.2",
+        "ocramius/proxy-manager": "^2.2",
         "friendsofphp/php-cs-fixer": "^2.4",
         "phpstan/phpstan": "^0.12"
     },
--- a/src/Compiler/Compiler.php
+++ b/src/Compiler/Compiler.php
@@ -21,7 +21,9 @@ use DI\Proxy\ProxyFactory;
 use function dirname;
 use function file_put_contents;
 use InvalidArgumentException;
-use Opis\Closure\SerializableClosure;
+use Laravel\SerializableClosure\SerializableClosure;
+use Laravel\SerializableClosure\Support\ReflectionClosure;
+use Opis\Closure\SerializableClosure as OpisSerializableClosure;
 use function rename;
 use function sprintf;
 use function tempnam;
@@ -401,8 +403,13 @@ PHP;
      */
     private function compileClosure(\Closure $closure) : string
     {
-        $wrapper = new SerializableClosure($closure);
-        $reflector = $wrapper->getReflector();
+        if (\PHP_VERSION_ID < 70400) {
+            $wrapper = new OpisSerializableClosure($closure);
+            $reflector = $wrapper->getReflector();
+        } else {
+            $wrapper = new SerializableClosure($closure);
+            $reflector = new ReflectionClosure($closure);
+        }
 
         if ($reflector->getUseVariables()) {
             throw new InvalidDefinition('Cannot compile closures which import variables using the `use` keyword');
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,10 +16,10 @@ jobs:
         timeout-minutes: 15
         strategy:
             matrix:
-                php: [ '7.2', '7.3', '7.4', '8.0' ]
+                php: [ '7.3', '7.4', '8.0', '8.1' ]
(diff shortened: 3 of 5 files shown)


src/Compiler/Compiler.php
```diff
@@ -402,7 +403,7 @@ private function isCompilable($value)
     private function compileClosure(\Closure $closure) : string
     {
         $wrapper = new SerializableClosure($closure);
-        $reflector = $wrapper->getReflector();
+        $reflector = new ReflectionClosure($closure);
```
shadowhand, line 406: Ah, it looks like #793 identified that `laravel/serializable-closure` only supports PHP >= 7.4, and this package supports PHP >= 7.2, so we would need to have both the Laravel package, and `opis/closure` installed. 😢 

Maybe it is time for PHP-DI to require supported PHP versions?

ldebrouwer: The alternative I guess is for the project to maintain a compatible fork of `opis/closure`. It'd be good to have some guidance of what the project maintainers envisage as a solution, so we can find a way forward.

shadowhand: I don't feel like that's necessary, given that Laravel has already relived this.

@mnapoli what is the likelihood that there can be a new major version of PHP-DI?

hemberger: Even if we don't get a major version bump for PHP-DI that only supports PHP 7.4+, I think that https://github.com/PHP-DI/PHP-DI/pull/793 is a fine interim solution to this problem. Yes, both closure packages would be installed, but in my opinion that's much more palatable than than getting spammed with deprecation messages. As it stands, PHP-DI is so noisy that I have to `@` suppress it, which makes it a liability in my codebase.

Hopefully we can get some movement on this issue from the maintainer, and thanks to those who have offered solutions! Appreciate you all!

mnapoli: Right, I've tried reopening #793 against master (now that there are stable versions), let's see if the tests pass: #810

mnapoli: OK so it seems this PR is the best approach.

`laravel/serializable-closure` requires PHP 7.3, not 7.4. So we could remove support from PHP 7.2 and release that as a new minor version. Composer will deal with PHP requirements.

Could you update everything (CI and composer.json requirement) to require PHP 7.3?

hemberger: Even if we bump the minimum requirement of PHP-DI to 7.3, we still need to use `opis/closure` for PHP 7.3. This is because `laravel/serializable-closure` really does require PHP 7.4, but allows it to be _installed_ (but not used) in PHP 7.3 (see https://github.com/laravel/serializable-closure/pull/31). Therefore, I think we really do need the solution provided by #793 / #810, but without PHP 7.2.

(And here is the requirement specifically in the code: https://github.com/laravel/serializable-closure/blob/bfbc6c8e85240d08d8324be3eae13b7563224a91/src/SerializableClosure.php#L28-L30)

shadowhand: @mnapoli there's another issue though, `ocramius/proxy-manager` can't be installed on PHP 8.1 because it version locks PHP as `~8.0.0` as of the latest version `2.14.1` and there is no (released) version that supports PHP 8.1. I can work around it, but I am mentioning it.

shadowhand:
I still don't love this. I hope that PHP-DI 7.0 can land soon.

src/Compiler/Compiler.php
mnapoli: > ocramius/proxy-manager can't be installed on PHP 8.1 because it version locks PHP as ~8.0.0 as of the latest version 2.14.1 and there is no (released) version that supports PHP 8.1.

Right yes, that's why I think it's OK if we don't run CI on 8.1 (unless you have a better idea in mind)

shadowhand:
@mnapoli test errors should be fixed now.

composer.json
```diff
@@ -25,17 +25,18 @@
...
-        "opis/closure": "^3.5.5"
+        "opis/closure": "^3.5.5",
+        "laravel/serializable-closure": "^1.0"
     },
     "require-dev": {
         "phpunit/phpunit": "^8.5|^9.0",
         "mnapoli/phpunit-easymock": "^1.2",
         "doctrine/annotations": "~1.2",
-        "ocramius/proxy-manager": "^2.0.2",
+        "ocramius/proxy-manager": "^2.2",
```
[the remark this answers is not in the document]
shadowhand:
```
ocramius/proxy-manager  2.2.0  requires  zendframework/zend-code (^3.3.0)  
```

shadowhand:
@mnapoli updated to fix newest build error.

composer.json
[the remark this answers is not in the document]
shadowhand: I don't know what to do about this test error, because [all versions of zendframework/zend-eventmanager][1] allow for version `^2.7 || ^3.0` of `zend-stdlib`, and newer versions of proxy-manager require `php ^7.4`. There is simply no way to resolve the build error while using `--prefer-lowest`.

[1]: https://packagist.org/packages/zendframework/zend-eventmanager#3.2.1

mnapoli:
Since we're stuck with PHP 7.3 support and this has been the case for a while, let's jump the gun and support PHP 7.4+ ?

I've implemented that in #811 based on this PR.

mnapoli:
Thanks!

netzhuffle:
@mnapoli PR in which direction: Removing the change log or bringing it back up to date?

mnapoli:
Ideally updating it as it's currently used in the website: https://php-di.org/change-log.html

PR #810, opened 2022-03-16
Makes serializable closure work with PHP 8.1

mnapoli:
Reopen https://github.com/PHP-DI/PHP-DI/pull/793

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,8 @@
         "psr/container": "^1.0",
         "php-di/invoker": "^2.0",
         "php-di/phpdoc-reader": "^2.0.1",
-        "opis/closure": "^3.5.5"
+        "opis/closure": "^3.5.5",
+        "laravel/serializable-closure": "^1.0"
     },
     "require-dev": {
         "phpunit/phpunit": "^8.5|^9.0",
--- a/src/Compiler/Compiler.php
+++ b/src/Compiler/Compiler.php
@@ -21,7 +21,9 @@ use DI\Proxy\ProxyFactory;
 use function dirname;
 use function file_put_contents;
 use InvalidArgumentException;
-use Opis\Closure\SerializableClosure;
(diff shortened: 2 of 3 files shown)


hemberger:
In my earlier comment about this being a workable solution (https://github.com/PHP-DI/PHP-DI/pull/808#discussion_r826616827), I forgot about the fundamental issue with the composer dependency requirements.

If there were an equivalent of the commandline `--ignore-platform-reqs=laravel/serializable-closure` that could be baked into composer.json, that would solve the problem, but I don't think such a thing exists. There is the [platform](https://getcomposer.org/doc/06-config.md#platform) option, which could fake a specific version of PHP for dependency resolution, but that seems like a dangerous hack because it applies to _all_ dependencies (not just `laravel/serializable-closure`).

Sadly, I'm all out of ideas here.

hemberger:
Maybe we can put `laravel/serializable-closure` in the [suggest](https://getcomposer.org/doc/04-schema.md#suggest) field and use it if it's installed?

Something like:
```json
{
    "suggest": {
        "laravel/serializable-closure": "Allows a closure interface that isn't deprecated in PHP 8.1+"
    }
}
```

```php
if (class_exists('Laravel\SerializableClosure\SerializableClosure')) {
    # use laravel/serializable-closure
} else {
    # use opis/closure
}
```

shadowhand:
@hemberger I like that solution as a minor release solution.

@mnapoli I understand that you probably don't want to rush a major version, but this seems like a very good time to drop support for unsupported PHP versions.

mnapoli:
🤦  I completely forgot about PHP 7.2 and the Composer requirement.

Yeah… I think we'll need to jump to PHP 7.3. We won't need a major release though.

plasid:
Yep pretty stupid - PHP7 is old news, now need to deal with this problem causing a MC system to fail.

PR #811, opened 2022-04-09
PHP 8.1 support by requiring PHP 7.4 or up

mnapoli:
Fixes https://github.com/PHP-DI/PHP-DI/issues/791

Apply #808 but up to PHP 7.4, we've been stuck for too long.

Let see if we can get the build to be green.

### DIFF
--- a/change-log.md
+++ b/change-log.md
@@ -1,5 +1,9 @@
 # Change log
 
+## 6.1.0
+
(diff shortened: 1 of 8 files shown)


change-log.md
```diff
@@ -1,5 +1,9 @@
 # Change log
 
+## 6.1.0
+
+- [#791](https://github.com/PHP-DI/PHP-DI/issues/791) Support PHP 8.1, remove support for PHP 7.2
```
shadowhand, line 5: `PHP < 7.4` (it was both 7.2 and 7.3)

PR #820, opened 2022-05-21
Document inability to autowire php-internal classes

### LINKED ISSUE #470: Internal classes cannot be instantiated
It's probably not a bug, but I think it worth noting (sorry if it already is): internal classes (like `Memcached`) cannot be instantiated as PHP-DI cannot determine their constructor parameters through reflection. I guess it also means that this is only true for classes with constructor parameters.

The solution is using a factory:

``` php
            \DI\factory(function($host, $port) {
                $memcached = new \Memcached();

                $memcached->addServer($host, $port);

                return $memcached;
            })
```

Again, I don't think this is a bug in PHP-DI, but it might worth to investigate what actually happens and write a few lines in the documentation.

SvenRtbg:
closes #470

I believe nobody came up with a proper solution in the discussion of that ticket, so the lowest common denominator is to document more explicitly what PHP-DI expects the user to do.

### DIFF
--- a/doc/autowiring.md
+++ b/doc/autowiring.md
@@ -70,3 +70,7 @@ class Database
 - `setLogger()` will not be called
 
 For those classes, you will need to use `DI\autowire()` in [PHP definitions](php-definitions.md) to declare explicitly what to inject.
+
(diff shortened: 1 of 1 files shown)


mnapoli:
Sorry I skipped this one, thank you!

PR #824, opened 2022-08-10
Unset from entriesBeingResolved when using set()

### LINKED ISSUE #823: Using set for entries being resolved
When using `$container->set('foo', 'bar')`, shouldn't the container unset `foo` from `entriesBeingResolved`?

Yes, I am dealing with circular dependencies. I have an object which is created in a service class, but the same service class is then using other services which the object is autowired into. I thought I could use `set` early on in my script to give the container a reference to my object, so other classes can resolve it.

oskarakerlund:
There is an array `entriesBeingResolved` being used to check for circular dependencies and infinite loops. However this array does not change when an entry is being explicitly set using `set()`. When an entry is given a value, it should be considered resolved.

The proposed change will unset the entry in `entriesBeingResolved` when using `set()`

Closes #823

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -301,6 +301,7 @@ class Container implements ContainerInterface, FactoryInterface, InvokerInterfac
         } else {
             $this->resolvedEntries[$name] = $value;
         }
+        unset($this->entriesBeingResolved[$name]);
     }
 
     /**


mnapoli:
Hi, I'm not sure I'm picturing the use case 100%. Could you add a test in the PR, that might help see a case where this happens.

oskarakerlund:
Sorry - I can no longer reproduce the issue. Closing issue and PR!

PR #827, opened 2022-08-22
ContainerBuilder.php: use generics for typehints

hemberger:
The type of container that is actually built may be more specialized
than the Container base class. As such, make sure we use the proper
return typehints.

I found this issue while testing extensions of the Container class to
support a resettable interface, as an alternative solution to #821.
This typing issue would arise for anyone trying to extend Container.

This issue also exists in the v7 branch (with a handful of fairly
straightforward conflicts from current master). If you would prefer
that I target the v7 branch instead, I'd be happy to make that change.

### DIFF
--- a/src/ContainerBuilder.php
+++ b/src/ContainerBuilder.php
@@ -31,12 +31,14 @@ use Psr\Container\ContainerInterface;
  *
  * @since  3.2
  * @author Matthieu Napoli <matthieu@mnapoli.fr>
+ *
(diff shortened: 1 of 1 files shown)


hemberger:
Superseded by https://github.com/PHP-DI/PHP-DI/pull/842, which uses a single template parameter instead of two. This allows the type to be inferred from the constructor arguments, which restores the default behavior in the case where no argument is specified.

PR #828, opened 2022-09-29
Add docs for cache namespace feature

rob4226:
I don't see the `cacheNamespace` argument mentioned anywhere in the docs for `enableDefinitionCache()`, that was added with #698. The only reason I found it is because I was having cache collisions and looked at the source code where I noticed this nice feature already exists.  So this commit adds a paragraph to the docs about the cache namespace feature.

### DIFF
--- a/doc/performances.md
+++ b/doc/performances.md
@@ -133,6 +133,11 @@ if (/* is production */) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


mnapoli:
Thanks!

PR #829, opened 2022-11-13
Upgrade PHPStan to 1.9

takaram:
This PR upgrades PHPStan 0.12 to 1.9.

Recently [PHPStan 0.12.100](https://github.com/phpstan/phpstan/releases/tag/0.12.100) was released, which prompts 0.12.x series users to upgrade. Now we should upgrade PHPStan too.

There are several errors increased by this upgrade, but I put them all to baseline for the time being.


mnapoli:
Thanks!

phpcs failures are unrelated and fail on master too 😢  I'll merge this!

Issue #834, opened 2023-01-08
[v7] Enabling static analysis on definitions


szepeviktor:
Using PHP generics (and array types) we can set key and value type of definitions throughout the code base.
Actually turning comments into code.
```php
/**
 * Keys are strings, values are `Definition` objects or null.
 *                    👇🏻
 * @var \ArrayIterator<string, \DI\Definition\Definition|null>`
 */
```

Now definitions may have a bug or two: their types are ignored.

@mnapoli Would you accept a PR?

szepeviktor:
This would make our PHPStan baseline an almost empty file!

mnapoli:
I would, but on the `v7` branch (because it will be merged to master next week). And v7 uses Psalm (a year ago it was ahead of phpstan).

szepeviktor:
All right.
[I](https://packagist.org/packages/szepeviktor/phpstan-wordpress) let you ride the `psalm` way.

PR #835, opened 2023-01-08
Remove outdated PHPStan ignores

szepeviktor:
All of them suppress non-existent errors.


mnapoli:
Unfortunately I think we need to close this one, see https://github.com/PHP-DI/PHP-DI/issues/834#issuecomment-1375239620

szepeviktor:
Done :)

PR #836, opened 2023-01-08
[v7] Fix changelog's file name

szepeviktor:
Just that.

https://en.wikipedia.org/wiki/Changelog


mnapoli:
We can't change the file name as this would break this URL: https://php-di.org/change-log.html

szepeviktor:
All right.
(HTTP redirection)

PR #838, opened 2023-01-08
Fix misspellings

szepeviktor:
Codespell could be [integrated](https://github.com/szepeviktor/debian-server-tools/blob/master/webserver/Continuous-integration-Continuous-delivery.yml#L189) in CI.

https://www.merriam-webster.com/dictionary/invocable

### DIFF
--- a/change-log.md
+++ b/change-log.md
@@ -210,8 +210,8 @@ Internal changes in case you were replacing/extending some parts:
 
 Read the [news entry](news/13-php-di-4-4-released.md).
 
-- [#185](https://github.com/PHP-DI/PHP-DI/issues/185) Support for invokable objects in `Container::call()`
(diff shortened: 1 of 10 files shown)


change-log.md
```diff
@@ -210,8 +210,8 @@ Internal changes in case you were replacing/extending some parts:
 
 Read the [news entry](news/13-php-di-4-4-released.md).
 
-- [#185](https://github.com/PHP-DI/PHP-DI/issues/185) Support for invokable objects in `Container::call()`
```
mnapoli, line 213: I'd rather keep `invokable` since both seem acceptable.

szepeviktor: All right.
I leave it up to you as I'm not a user just a person sending random PR-s.

Issue #840, opened 2023-01-09
Supporting named arguments in PHP-DI 7 


ldebrouwer:
I took the first release candidate for PHP-DI 7 for a spin on a large project, and it was a very smooth upgrade process, but I did notice a slight regression in terms of developer experience.

In PHP-DI 6 we could easily use the following;

```
    App\Service::class => autowire()
        ->constructor(
            enabled: fn() => Config::ENABLED,
            URL: fn() => Config::REDIRECT_URI,
            clientID: fn() => Config::CLIENT_ID,
            clientSecret: fn() => Config::CLIENT_SECRET,
        ),
 ```
 As you can see we could easily use PHP 8's named arguments, and it would work as intended.
 
 And with 7.0.0-rc1 we needed to update this to something like this;
 
```
    App\Service::class => autowire()
        ->constructorParameter('enabled', fn() => Config::ENABLED)
        ->constructorParameter('URL', fn() => Config::REDIRECT_URI)
        ->constructorParameter('clientID', fn() => Config::CLIENT_ID)
        ->constructorParameter('clientSecret', fn() => Config::CLIENT_SECRET),
 ```
 
 It's not a massive issue per se, but it _was_ really nice to be able to use named arguments.

mnapoli:
Thanks for the report.

Could you clarify what happens if you use named arguments with PHP-DI 7?

ldebrouwer:
It _works_, but PHPStan throws an error like the one below, because of the typehint;
```
Parameter ...$parameters of method DI\Definition\Helper\CreateDefinitionHelper::constructor() expects array, Closure given.
```

mnapoli:
Ah ok so it's just a phpstan warning. Feel free to send a PR if you find out how to solve this.

PR #841, opened 2023-01-09
Correct type hints to prevent tooling like PHPStan from raising issues.

ldebrouwer:
This PR allows people who use PHPStan in their tool chain to use named arguments when calling the `method` and `constructor` methods of `CreateDefinitionHelper`.

See https://github.com/PHP-DI/PHP-DI/issues/840

### DIFF
--- a/src/Definition/Helper/CreateDefinitionHelper.php
+++ b/src/Definition/Helper/CreateDefinitionHelper.php
@@ -104,7 +104,7 @@ class CreateDefinitionHelper implements DefinitionHelper
      * Can be used multiple times to declare multiple calls.
      *
      * @param string $method       Name of the method to call.
-     * @param mixed $parameters Parameters to use for calling the method.
+     * @param mixed ...$parameters Parameters to use for calling the method.
      *
      * @return $this
      */
--- a/src/Definition/Helper/CreateDefinitionHelper.php
+++ b/src/Definition/Helper/CreateDefinitionHelper.php
@@ -68,7 +68,7 @@ class CreateDefinitionHelper implements DefinitionHelper
      * This method takes a variable number of arguments, example:
      *     ->constructor($param1, $param2, $param3)
      *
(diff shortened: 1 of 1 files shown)


src/Definition/Helper/CreateDefinitionHelper.php
```diff
@@ -68,7 +68,7 @@ public function lazy() : self
      * This method takes a variable number of arguments, example:
      *     ->constructor($param1, $param2, $param3)
      *
-     * @param mixed[] $parameters Parameters to use for calling the constructor of the class.
+     * @param mixed $parameters Parameters to use for calling the constructor of the class.
```
jdreesen, line 71: I think it should be like this for [variadic functions](https://phpstan.org/writing-php-code/phpdocs-basics#variadic-functions):

```suggestion
     * @param mixed ...$parameters Parameters to use for calling the constructor of the class.
```

src/Definition/Helper/CreateDefinitionHelper.php
```diff
@@ -104,7 +104,7 @@ public function property(string $property, mixed $value) : self
      * Can be used multiple times to declare multiple calls.
      *
      * @param string $method       Name of the method to call.
-     * @param mixed[] $parameters Parameters to use for calling the method.
+     * @param mixed $parameters Parameters to use for calling the method.
```
jdreesen, line 107:
```suggestion
     * @param mixed ...$parameters Parameters to use for calling the method.
```

src/Definition/Helper/CreateDefinitionHelper.php
ldebrouwer: Let me confirm that that works, and I'll update it accordingly.

ldebrouwer:
As per @jdreesen's suggestions I prefixed the hints with `...` to denote that we're dealing with variadic functions/arguments.

mnapoli:
Thanks!

PR #843, opened 2023-01-13
CompiledContainer: match Container interface

hemberger:
In 9df6976824d08f, the Container interface was modified so that the `has` and `get` methods strictly type their parameters to conform to psr/container 1.1.

This modifies CompiledContainer to do exactly the same.


mnapoli:
Good catch thank you!

Issue #845, opened 2023-01-27
declare(strict_types=1) in vendor/php-di/php-di/src/Compiler/Template.php


rustamr1991:
When add declare(strict_types=1) everywhere in [issue](https://github.com/PHP-DI/PHP-DI/issues/496) forgot to add Template.php
The problem occurs when using the enableCompilation method

PR #846, opened 2023-01-27
add strict types in Template.php

rustamr1991:
When add declare(strict_types=1) everywhere in https://github.com/PHP-DI/PHP-DI/issues/496 forgot to add Template.php
The problem occurs when using the enableCompilation method
fix issue https://github.com/PHP-DI/PHP-DI/issues/845

### DIFF
--- a/src/Compiler/Compiler.php
+++ b/src/Compiler/Compiler.php
@@ -155,7 +155,7 @@ class Compiler
         require __DIR__ . '/Template.php';
         $fileContent = ob_get_clean();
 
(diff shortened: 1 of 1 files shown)


mnapoli:
Tests are failing for some reason. Can you look into it?

alexandrmazur96:
Tests fails because `declare(...)` without trailing `;`, so such compiled container will not work in any case.

Also, this PR duplicates #864 (or vice versa). Anyway, these changes should not be merged (I described why in #864)

mnapoli:
Got it, thanks!

PR #849, opened 2023-02-07
ContainerBuilder: fallback typehints for PhpStorm

### LINKED ISSUE #844: 7.x IDE Integration broken?
Hi :wave:,

I'm doing some tests with version 7.x before using it more globally.
I was using the [.phpstorm.meta.php trick](https://php-di.org/doc/ide-integration.html#metadata-file) to help PhpStorm resolving correctly the classes fetch via container. It doesn't work anymore.

I see that PhpDoc has changed, and, If i'm correct, it uses now [psalm](https://psalm.dev)?

```php
    /**
     * Build and return a container.
     *
     * @return ContainerClass
     */
    public function build()
```

I'm using this trick right now.

```php
/** @var Container $container */
```

Is it a "misuses" of psalm from PHP-DI side, or a bug I will report to PhpStorm?

Thanks.

Regards.

hemberger:
Fixes #844.

PhpStorm currently lacks the ability to infer a template type from a `class-string` constructor argument (see
https://youtrack.jetbrains.com/issue/WI-65811/Infer-type-from-constructors-class-string-parameter). As a result of #842 (v7.0.1), this requires PhpStorm users to manually typehint their ContainerBuilder instances, e.g.:

    /** @var ContainerBuilder<Container> $builder */

To avoid this, we only use the proper templated typehints with Psalm- and PHPStan-specific annotations, and leave the non-templated typehints as a fallback. Note that the fallbacks will not give the correct type when using a non-default Container class.

@Gounlaf While this should in theory recover the pre-7.0.1 behavior for PhpStorm, I'm not able to test this. Could you please check if it fixes your issue? Many thanks!

### DIFF
--- a/src/ContainerBuilder.php
+++ b/src/ContainerBuilder.php
@@ -32,19 +32,21 @@ use Psr\Container\ContainerInterface;
  * @since  3.2
  * @author Matthieu Napoli <matthieu@mnapoli.fr>
  *
- * @template ContainerClass of Container
+ * @psalm-template ContainerClass of Container
  */
 class ContainerBuilder
 {
     /**
      * Name of the container class, used to create the container.
-     * @var class-string<ContainerClass>
+     * @var class-string<Container>
+     * @psalm-var class-string<ContainerClass>
      */
     private string $containerClass;
 
     /**
      * Name of the container parent class, used on compiled container.
-     * @var class-string<ContainerClass>
+     * @var class-string<Container>
+     * @psalm-var class-string<ContainerClass>
      */
     private string $containerParentClass;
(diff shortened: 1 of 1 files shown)


mnapoli:
Thanks!

These are a lot of annotations, couldn't we limit ourselves to reverting #842 for now? Maintaining all those annotations is not something I am looking for tbh :/

jdreesen:
I don't think we need both `@psalm-*` and `@phpstan-*` as each tool can understand both.

mnapoli:
@hemberger yes that's better, thank you!

PR #850, opened 2023-03-08
Fix `useAttributes` missing attribute

lcharette:
Documentation is missing an argument. See : https://php-di.org/doc/attributes.html#setup


mnapoli:
Thanks! Funny I thought this parameter was optional, it probably should be worth making it optional at some point. But in the meantime let's merge that!

lcharette:
Making the argument optional would make more sense since the option is "false" by default.

PR #856, opened 2023-06-15
Fix injection of readonly promoted properties

### LINKED ISSUE #853: #[Inject] on readonly constructor-promoted properties
Injection currently only works when the Inject attribute is used as a method attribute on the constructor. 

When the Inject attribute is used on the readonly constructor-promoted property, PHP-DI attempts to set the value through ```\DI\Definition\Resolver\ObjectCreator::setPrivatePropertyValue()``` but fails because ```setValue()``` can't be called on a readonly property.

apeschar:
When `#[Inject]` annotations are used on promoted properties, PHP-DI quietly injects the value twice: once as a constructor argument, and once as an injected property. This is because PHP [replicates the attribute on the constructor argument to the property](https://www.php.net/manual/en/language.oop5.decon.php#language.oop5.decon.constructor.promotion).

Generally, the second unneeded injection goes unnoticed, but when `readonly` properties are used, the property injection fails:

    Error: Cannot modify readonly property DI\Test\IntegrationTest\Attributes\PromotedReadonlyProperty::$promotedProperty

This PR excludes promoted properties from injection. Promoted properties should never be injected because they will already have been assigned a value in the constructor.

This fixes #853.

### DIFF
--- a/src/Definition/Source/AttributeBasedAutowiring.php
+++ b/src/Definition/Source/AttributeBasedAutowiring.php
@@ -77,9 +77,6 @@ class AttributeBasedAutowiring implements DefinitionSource, Autowiring
     private function readProperties(ReflectionClass $class, ObjectDefinition $definition) : void
     {
         foreach ($class->getProperties() as $property) {
-            if ($property->isStatic()) {
-                continue;
-            }
             $this->readProperty($property, $definition);
         }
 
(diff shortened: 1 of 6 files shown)


mnapoli:
That is looking brilliant! Thanks for the very clean implementation 🙌

I fixed the build on master with #857, can you merge master in this branch, or rebase?

apeschar:
Rebased. ✅  You just have to approve the workflow.

PR #859, opened 2023-08-08
Enum value definitions

zorac:
PHP 8.1 added native enumerations, but currently using an enum case as a value definition generates an error when attempting to compile the container: `An object was found but objects cannot be compiled`. This change simply allows objects which are enum cases to be compiled. I've also added tests for enum cases as value definitions.

### DIFF
--- a/.php-cs-fixer.php
+++ b/.php-cs-fixer.php
@@ -51,7 +51,8 @@ return $config->setRules([
     'single_line_throw' => false,
     'php_unit_method_casing' => false,
     'blank_line_between_import_groups' => false,
-        'global_namespace_import' => false,
+    'global_namespace_import' => false,
+    'nullable_type_declaration_for_default_null_value' => false,
 ])
     ->setRiskyAllowed(true)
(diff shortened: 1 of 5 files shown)


mnapoli:
Great PR, thanks!

There seem to be some failures with some jobs, but happy to merge once that's fixed 👍

zorac:
OK, as Psalm is running on PHP 8.0, it doesn't like the reference to the PHP 8.1-only `UnitEnum` interface, even after adding a version check, so I've added a `@psalm-suppress` for that.

The coding standards fixer is complaining about something in a file I didn't touch, so possibly something's changed in their rules? Rather than changing a user-facing class, I've disabled the offending rule in `.php-cs-fixer.php` which allows it to run without changing things either way. (It doesn't help that the documentation for this rule uses conflicting names, and doesn't appear to match the observed behaviour... https://cs.symfony.com/doc/rules/function_notation/nullable_type_declaration_for_default_null_value.html)

mnapoli:
Aweeeesome!

PR #860, opened 2023-08-08
Fix public readonly property injection

kocsismate:
The compiler tries to initialize all public properties directly inside the compiled container, however readonly properties do not support writing from an outside scope.

Additionally, I noticed a small room for a performance improvement: `ReflectionProperty::setAccessible()` doesn't have to be invoked as of PHP 8.1 (https://wiki.php.net/rfc/make-reflection-setaccessible-no-op), so I fixed it as well in a separate commit.


mnapoli (approved):
That's brilliant! The CS checks are failing, but once fixed I'd be happy to merge!

kocsismate:
Done! The one remaining CS issue seems unrelated to my changes 🤔

mnapoli:
Got it, thanks!

PR #862, opened 2023-08-31
Fix wildcard definition global namespace matching any namespace

### LINKED ISSUE #861: Wildcard definition for class in global namespace matches any namespace
The project I am working on has quite some classes defined in the global namespace. 
When adding a wildcard definition for these classes, not only those classes are matched, but also matching class names in any namespace.
This is quite unexpected, since the documentation states:
> the wildcard does not match across namespaces

For example:

```
return [
    'Job*' => DI\autowire()->method('foo', 'bar'),
];
```
will call `foo('bar')` on class `JobDoSomething`, but also on class `SomeNamespace\JobScheduler`, which is not in the global namespace.

I am unfortunately on version 6.3.5, because PHP 7.2 ... But the code for matching the wildcard (regex) looks to be equal in both 6.4 and master.

Micr0mega:
Fixes #861

### DIFF
--- a/src/Definition/Source/DefinitionArray.php
+++ b/src/Definition/Source/DefinitionArray.php
@@ -86,7 +86,7 @@ class DefinitionArray implements DefinitionSource, MutableDefinitionSource
         foreach ($this->wildcardDefinitions as $key => $definition) {
             // Turn the pattern into a regex
             $key = preg_quote($key, '#');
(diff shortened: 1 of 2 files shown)


mnapoli:
Thanks! There seems to be slight adjustments needed to code formatting, and then I can merge 👍

mnapoli:
Oh nevermind it's not related to the PR!

PR #864, opened 2023-10-01
Add declare_strict() in compiled container

krollobrocket:
Add declare(strict_types=1) in the file generated when compiling the container.

### DIFF
--- a/src/Compiler/Compiler.php
+++ b/src/Compiler/Compiler.php
@@ -155,7 +155,7 @@ class Compiler
         require __DIR__ . '/Template.php';
         $fileContent = ob_get_clean();
 
-        $fileContent = "<?php\n" . $fileContent;
+        $fileContent = "<?php declare(strict_types=1);\n" . $fileContent;
 
         $this->createCompilationDirectory(dirname($fileName));
         $this->writeFileAtomic($fileName, $fileContent);


mnapoli:
I wonder if this can have side-effects 🤔

We're embedding user code in there, what if they are not using the strict mode? 🤔

alexandrmazur96:
The following definition:

```php
'test' => static function (): string {
    $fn = static fn(int $a): int => $a + 1;
    $fn(1.25);

    return 'test';
}
```

produces the following code in the compiled container:

```php
<?php

/**
 * This class has been auto-generated by PHP-DI.
 */
class CompiledContainer extends DI\CompiledContainer{
    const METHOD_MAPPING = array (
  'test' => 'get1',
);

    protected function get1()
    {
        return $this->resolveFactory(static function (): string {
        $fn = static fn(int $a): int => $a + 1;
        $fn(1.25);

        return 'test';
    }, 'test');
    }

}
```

that successfully executed:

```php
<?php

require_once __DIR__ . '/vendor/autoload.php';
require_once __DIR__ . '/var/cache/di-container/CompiledContainer.php';

$c = new CompiledContainer();

var_dump($c->get('test'));
```
>  string(4) "test"

But if you add `declare(strict_types=1);` you, obviously, got an error:

> PHP Fatal error:  Uncaught TypeError: CompiledContainer::{closure}(): Argument #1 ($a) must be of type int, float given, called in /path/to/CompiledContainer.php on line 17 and defined in /path/to/CompiledContainer.php:16

So, I think these changes can not be merged.

juliangut:
https://3v4l.org/Cn7JP

I'm not going to say using strict_types is the way to go or not, or that passing a float to a function that defines its argument as integer is wrong, because this is PHP and we all know how it quirky can be

But I can see a point on allowing the user to decide whether to add it or not would be a solution, in the same way you decide if you include it in your source files. Does it mean yet another argument for enableCompilation?

mnapoli:
Right, thanks for confirming. TBH I don't want to take the risk, nor add yet another option for this. Let's close for now.

alexandrmazur96:
@juliangut, what potential benefits can strict types bring? I understand if we discuss the code outside the library, but here?

juliangut:
Hi

I think strict_types is already very explicit in what benefits it brings to the table.

If we have a look at the issue recently closed (I'm on the move and can't look it up) passing a float to a function declaring as an int parameter IS an error. But then PHP replies at you with type coercion and transparently transforms that float to int. With strict types it is an error straight

Why is this not the standard behavior? Because it'd had broken all PHP ecosystem, but nowadays I see it as the "normal" way of developing in PHP, do not know how someone relies in PHP magically transforming your types, it is a sloppy or hacked way of writing PHP today (if you are already using types)

Adding it to the container from my point of view would be correct but a developer decision. If I rely on strict types anywhere why shouldn't I expect PHP not to trigger an exception if I return an object not defined in the return type? Or use the wrong type for a constructor? or ..?. Given mostly the container is created by phpdi it is already free from error, so it comes to anonymous function factories only. As I mentioned it is another option for compilation or new parameter, a bit of a mess

Issue #866, opened 2023-10-27
WriteProxiesToFile bug


radar3301:
**Version 7 introduced a bug with `ContainerBuilder::writeProxiesToFile()`.**

Previous versions of ProxyFactory had a `bool $writeProxiesToFile` property.
Since 7.0, these checks were updated from `if ($this->writeProxiesToFile) ...` to `if ($this->proxyDirectory) ...`.
No issues here.

Prior to 7.0, ProxyFactory was created in ContainerBuilder by `new ProxyFactory($this->writeProxiesToFile, $this->proxyDirectory)`.
Since 7.0, ProxyFactory is created in ContainerBuilder by `new ProxyFactory($this->proxyDirectory)`.
No issues here.

Until you get to `ContainerBuilder::writeProxiesToFile(bool $writeToFile, string $proxyDirectory = null)`:
The `writeProxiesToFile` parameter essentially has no effect if `proxyDirectory` has anything other than null.

In other words, `writeProxiesToFile(false, 'tmp/proxies')` has exactly the same effect as `writeProxiesToFile(true, 'tmp/proxies')` or even
```php
$builder->writeProxiesToFile(true, 'tmp/proxies');
/* other code */
if ($dontWriteProxiesAnymore) {
    $builder->writeProxiesToFile(false);
}
```
or
```php
$builder->writeProxiesToFile($resultOfSomeLogic, $someProxyDirectory);
```

Therefore, to fix this bug, I believe [this line](https://github.com/PHP-DI/PHP-DI/blob/9ea40a5a6970bf1ca5cbe148bc16cbad6ca3db6c/src/ContainerBuilder.php#L250) should be updated to
```php
$this->proxyDirectory = $writeToFile ? $proxyDirectory : null;
```

Thoughts?

mnapoli:
Makes sense, thanks for spotting that! I'd be happy to merge a PR with a fix and a test covering the regression!

PR #868, opened 2023-10-31
Fix write proxies regression

radar3301:
Fixes the `ContainerBuilder->writeToProxies` regression introduced in 7.0 (see #866).


mnapoli:
Thanks a lot!

PR #869, opened 2023-12-15
Make Container return template type with conditional return type

zonuexe:
Generally when we pass a class name to a container we expect only objects of that class to be returned, not mixed.

In particular, PHPStan normalizes the `T|mixed` type to `mixed`, so the old type would not work.

Conditional return types notation is supported by both Psalm and PHPStan and provides a more appropriate type for both.

- https://psalm.dev/docs/annotating_code/type_syntax/conditional_types/
- https://phpstan.org/blog/phpstan-1-6-0-with-conditional-return-types

I've been using this patch for over a year now and it's working fine with PHPStan in our application.

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -118,6 +118,7 @@ class Container implements ContainerInterface, FactoryInterface, InvokerInterfac
      * @param string|class-string<T> $id Entry name or a class name.
      *
      * @return mixed|T
+     * @psalm-return ($id is class-string<T> ? T : mixed)
      * @throws DependencyException Error while resolving the entry.
      * @throws NotFoundException No entry found for the given name.
      */
@@ -165,6 +166,7 @@ class Container implements ContainerInterface, FactoryInterface, InvokerInterfac
      *                                           array will be resolved using the container.
      *
      * @return mixed|T
+     * @psalm-return ($id is class-string<T> ? T : mixed)
      * @throws InvalidArgumentException The name parameter must be of type string.
      * @throws DependencyException Error while resolving the entry.
      * @throws NotFoundException No entry found for the given name.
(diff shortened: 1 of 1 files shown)


mnapoli:
That looks very interesting, thanks!

2 questions:

- can you confirm that if I retrieve a non-class (e.g. a string parameter like `$container->get('db.host')`) then the inferred return type will be `mixed`?
- should this be annotated on PSR-11's ContainerInterface instead?

zonuexe:
> can you confirm that if I retrieve a non-class (e.g. a string parameter like $container->get('db.host')) then the inferred return type will be mixed?

You can check that these types are accepted with code like the following:

 * https://phpstan.org/r/48eaf957-0668-4695-aa6a-17fbcda5087a
 * https://psalm.dev/r/9c82ba9298

> should this be annotated on PSR-11's ContainerInterface instead?

Logically it is possible, and in our application we add the same type to PSR-11 and PSR-7 ServerRequest with [composer-patches](https://github.com/cweagans/composer-patches).

However, the PSR documentation does not define the semantics of these `$id`s, so it appears that adding them to the interface will require modifying PSR. In my opinion, it is enough to extend types at the responsibility of libraries and applications rather than recommendations.

mnapoli:
🤔 right. Unfortunately, I think the same logic applies to this class and PSR-11. I'd rather be consistent with PSR-11.

MarcHagen:
Found this PR, and we are having the same "problem" (first world problems right 😅).

_Don't know if the playground could be better, but it reflects the error given ignore others_
https://phpstan.org/r/5ed37a8b-e8d6-423e-a78d-69ccdc5ce1af
```
Property TestClass::$app (App) does not accept mixed.
```

I my opinion the package has changed the PSR-11 PHPDoc to add the generics in PHP-DI. 
If you think if should be consistent with the PSR-11 then remove the generics.

mnapoli:
Closing per comment above.

staabm:
@zonuexe @MarcHagen I guess this PHPStan extension implements what you try to achieve 

https://github.com/bnf/phpstan-psr-container

PR #870, opened 2023-12-17
Update tests for phpunit 10 compatibility

williamdes:
I propose this patch because it landed into Debian as https://sources.debian.org/src/php-di/7.0.5-1/debian/patches/0002-Ajust-tests-for-phpunit-10.patch/

I understand that currently phpunit 10 can not be run because of `mnapoli/phpunit-easymock` but all the content still makes sense. As providers should have been static since quite some phpunit versions. Only enforced now. (we run the test suite with phpunit 10 at Debian it works great)

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -17,6 +17,7 @@
         </testsuite>
         <testsuite name="integration">
             <directory>./tests/IntegrationTest/</directory>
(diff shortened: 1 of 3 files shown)


williamdes:
Hello @mnapoli 
Would you mind having a look to this one please ?

mnapoli:
Sorry that slipped through my list, sounds great thanks!

PR #879, opened 2024-02-02
adds windows-latest to github action tests and corrects PHP_EOL to \n…

marcing:
+ bumped actions to v4 to avoid warnings

Corrected PHP_EOL to \n, because [of ](https://github.com/PHP-DI/PHP-DI/pull/276) was not a proper fix for that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,19 +11,20 @@ on:
 jobs:
 
     tests:
-        name: Tests - PHP ${{ matrix.php }} ${{ matrix.dependency-version }}
-        runs-on: ubuntu-latest
+        name: Tests ${{ matrix.os }} - PHP ${{ matrix.php }}
         timeout-minutes: 15
         strategy:
             matrix:
+                os: [ubuntu-latest, windows-latest]
                 php: [ '8.0', '8.1', '8.2', '8.3' ]
-                dependency-version: [ '' ]
-                include:
-                    -   php: '8.0'
-                        dependency-version: '--prefer-lowest'
+        runs-on: ${{ matrix.os }}        
         steps:
+            -   name: Set git to use LF
+                run: |
+                    git config --global core.autocrlf input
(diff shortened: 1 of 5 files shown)


mnapoli:
Hi, could you explain a bit more the changes? Using that constant is usually considered a good practice. I'm surprised you want to revert a change that was done 8 years ago.

SvenRtbg:
I don't know about the motivation, however the constant is platform dependent, and that can result in all kinds of "unreliable line endings" issues if you work cross-platform in any way. If Windows is used to create a file, and Linux consumes it or vice versa, relying on PHP_EOL pretends to deal with that issue, but instead creates it.

So in essence, I would not call it "best practice" to use it. It may be useful in certain situations where outputting to some platform-dependent channel (like the console) is better done using PHP_EOL - on the other side the Windows port of PHP really goes to great lengths making the Windows behind it invisible, e.g. paths do not need to use backslashes, they can use slashes on Windows.

Still I wonder what issue this would resolve, and I am waiting to read an answer as well.

mnapoli:
Yeah you make a good point. I'm more in the camp of "what is wrong now, and what is fixed with this PR" because it's a significant change.

marcing:
@mnapoli @SvenRtbg 

The issue here was that we replaced \n with PHP_EOL in methods, which are used to compare strings present in unit test source files. Variables with strings containing line endings should not depend on platform (contrary to some file system operations). The coding standard on the project uses \n as EOL, that is why we should not compare it against \r\n (which is the value of PHP_EOL on Windows). The original pull request from 8 years ago passed tests only because of local GIT settings of the author, which converted \n in source code to \r\n when checking out, which was essentially an alteration of the source code files. This pull request fixes that, because the tests did not pass on Windows since then. New github action was added which runs the tests on windows, additionally with ```git config --global core.eol lf``` configuration to make sure same line endings are the same as the ones in source files. 

Let me know what you think guys, I'm open for discussing it further. Thank you.

PR #882, opened 2024-02-28
Added "Interface" suffix to interface names

mdmm-git:
PSR Naming Conventions

1. Interfaces MUST be suffixed by Interface: e.g. Psr\Foo\BarInterface.


jdreesen:
Might be a tiiiny little BC break :laughing:

mdmm-git:
> Might be a tiiiny little BC break 😆

Yeah, that's true😅
But in theory it's more correct and the code reads better. It just fits one of the newer major versions. I personally like the IBar variant better than BarInterface, but as long as it's accepted and most people use it, why not

Issue #881, opened 2024-02-25
method injection only passes the first parameter if the method uses a splat for its parameter


joachim-n:
If I do:

```
  'letters' => DI\create(Letters::class)
    ->method(
      'setLetters',
      DI\get('alpha'),
      DI\get('beta'),
    ),
```

and 

```
  public function setLetters($alpha, $beta) {
```

then both services are passed in. But if I change the method signature to this:

```
  public function setLetters(...$letters) {
```

Then $letters is an array with only the first service.

Full test code:

```
<?php

require 'vendor/autoload.php';

$container = new DI\Container([
  'alpha'   => DI\create(Alpha::class),
  'beta'    => DI\create(Beta::class),
  'letters' => DI\create(Letters::class)
    ->method(
      'setLettersList',
      DI\get('alpha'),
      DI\get('beta'),
    )
    ->method(
      'setLettersSplat',
      DI\get('alpha'),
      DI\get('beta'),
    ),
]);

class Alpha {}
class Beta {}

class Letters {
  public function setLettersList($alpha, $beta) {
    dump(__METHOD__);
    dump(func_get_args());
  }

  public function setLettersSplat(...$letters) {
    // Should also get all the letters, but only gets the first one.
    dump(__METHOD__);
    dump(func_get_args());
  }
}

$letters = $container->get('letters');
```

joachim-n:
ObjectCreator::injectMethodsAndProperties() seems to be correctly getting $args in both cases, which makes it look like the problem is when the method is invoked here:

```
            $methodReflection->invokeArgs($object, $args);
```

But doing that directly on the objects works properly:

```
$letters = new Letters();
$methodReflection = new \ReflectionMethod($letters, 'setLettersSplat');
$args = [1, 2];
$methodReflection->invokeArgs($letters, $args);
```

joachim-n:
My mistake -- I got confused with $args in the loop.

This line doesn't get the complete array but only the first item:

```
            $args = $this->parameterResolver->resolveParameters($methodInjection, $methodReflection);
```

serge-kvashnin:
@joachim-n The PR I created should fix the issue you mentioned: https://github.com/PHP-DI/PHP-DI/pull/883

Issue #802, opened 2022-01-20
Dealing with splat operator in a Factory


emulgeator:
My question would be if it's possible somehow to tell `\DI\FactoryInterface::make` how to handle a variadic constructor.

So imagine the following class:
```
class Test
{
    public function __construct(SomeHelper $someHelper, int ...$extraParameters)
    {
    }
}

class Factory
{
    private FactoryInterface $diFactory;

    public function __construct(FactoryInterface $diFactory)
    {
        $this->diFactory = $diFactory;
    }

    public function getTest(int ...$extraParameters): Test
    {
        return $this->diFactory->make(
            Test::class,
            [
                ...$extraParameters,
            ]
        );
    }
}
```

My factory method obviously wont work this way, and now I can only instantiate the Test class by hand. Is there a way currently to handle this case or if not, would it be possible to support it somehow?

mnapoli:
🤔  that's a good question, honestly I'm not 100% sure. I would personally test passing the `extraParameters` parameter as an array, but there's 50% chances it works 😄

emulgeator:
Yeah I gave it a try already with eyes closed. Sadly it didn't work

mnapoli:
Would it work with

```php
final class AggregateThingService implements ThingServiceProviderInterface
{
    public function __construct(
        private readonly ThingServiceProviderInterface $defaultProvider,
        ThingServiceProviderInterface ...$additionalProviders = null,
    ) {
    }
}
```

? (i.e. suggesting to PHP-DI that these extra parameters are optional)

But maybe I'm just off… Does the splat operator make the parameter optional already?

PR #883, opened 2024-04-16
Resolve variadic parameters.

serge-kvashnin:
`ParameterResolver::resolveParameters()` was iterating through the reflection parameters and fetching appropriate values from either named parameters or definition parameters. This meant the number of resolved arguments matched the number of reflection parameters, even if a parameter was variadic. This change adds a check for variadic parameters and adds the rest of the parameters to the `args` array accordingly.

### DIFF
--- a/src/Definition/Resolver/ParameterResolver.php
+++ b/src/Definition/Resolver/ParameterResolver.php
@@ -48,6 +48,26 @@ class ParameterResolver
                 // Look in the $parameters array
                 $value = &$parameters[$parameter->getName()];
             } elseif (array_key_exists($index, $definitionParameters)) {
+                if ($parameter->isVariadic()) {
+                    $args = array_merge(
+                        $args,
+                        // Resolve definitions, if any
+                        array_map(
+                            function ($arg) {
+                                if ($arg instanceof Definition) {
+                                    return $this->definitionResolver->resolve($arg);
+                                }
+
+                                return $arg;
+                            },
+                            // Parameters that are left
+                            array_slice($definitionParameters, $index)
+                        )
+                    );
+
+                    break;
+                }
+
                 // Look in the definition
                 $value = &$definitionParameters[$index];
             } else {
--- a/tests/UnitTest/Definition/Resolver/Fixture/FixtureClass.php
+++ b/tests/UnitTest/Definition/Resolver/Fixture/FixtureClass.php
@@ -10,6 +10,7 @@ class FixtureClass
     public $constructorParam1;
     public $methodParam1;
     public $methodParam2;
+    public $methodParam3;
 
     public function __construct($param1)
     {
@@ -24,5 +25,10 @@ class FixtureClass
     public function methodDefaultValue($param = 'defaultValue')
     {
         $this->methodParam2 = $param;
+    }
+
+    public function methodWithVariadicParameter($param1, ...$param2)
+    {
+        $this->methodParam3 = [$param1, ...$param2];
     }
 }
--- a/tests/UnitTest/Definition/Resolver/ObjectCreatorTest.php
+++ b/tests/UnitTest/Definition/Resolver/ObjectCreatorTest.php
@@ -54,6 +54,18 @@ class ObjectCreatorTest extends TestCase
         $this->assertEquals('value3', $object->methodParam1);
     }
 
+    public function testResolveWithVariadicParameter()
+    {
+        $definition = new ObjectDefinition(FixtureClass::class);
+        $definition->setConstructorInjection(MethodInjection::constructor(['value1']));
+        $definition->addMethodInjection(new MethodInjection('methodWithVariadicParameter', ['value1', 'value2', 'value3']));
+
+        $object = $this->resolver->resolve($definition);
+
+        $this->assertInstanceOf(FixtureClass::class, $object);
+        $this->assertEquals(['value1', 'value2', 'value3'], $object->methodParam3);
+    }
+
     public function testResolveNoConstructorClass()
     {
         $definition = new ObjectDefinition(NoConstructor::class);


serge-kvashnin:
Solves https://github.com/PHP-DI/PHP-DI/issues/881

mnapoli:
Any significant feature addition like this requires a new integration test, can you add it? I'm not sure how it behaves with the compiled container.

serge-kvashnin:
@mnapoli You're right. Added test and handling of variadic parameter for `ObjectCreationCompiler`.

.php-cs-fixer.php
```diff
@@ -58,6 +58,7 @@
     'blank_line_between_import_groups' => false,
     'global_namespace_import' => false,
     'nullable_type_declaration_for_default_null_value' => false,
+    'nullable_type_declaration' => false,
```
mnapoli, line 61: Why this change?

tests/IntegrationTest/CompiledContainerTest.php
```diff
@@ -177,6 +177,41 @@ public function proxy_classes_can_be_pregenerated_at_compile_time()
         $this->assertEquals(2, $countProxyClasses);
     }
 
+    /**
+     * @test
+     * @see https://github.com/PHP-DI/PHP-DI/issues/881
+     */
+    public function works_with_variadic_parameters()
```
mnapoli, line 184: Why not add this to the base containter test class? This isn't specific to the compiled container, variadic parametrs should be supported (and so tested) on both compiled and non-compiled containers.

serge-kvashnin: Created separate test case along with other issue cases, for both compiled and non-compiled containers.

.php-cs-fixer.php
mnapoli: Oh you mean that fixes master?

A bit more clarity would help here 🙂

tests/IntegrationTest/Issues/Issue881Test.php
```diff
@@ -0,0 +1,78 @@
...
+use DI\Test\IntegrationTest\Issues\Issue881\A;
+use DI\Test\IntegrationTest\Issues\Issue881\B;
+use DI\Test\IntegrationTest\Issues\Issue881\C;
+use DI\Test\IntegrationTest\Issues\Issue881\ClassWithVariadicParameter;
+use function DI\create;
+use function DI\get;
+
+final class Issue881Test extends BaseContainerTest
+{
+    public function testContainerWithVariadicParameters(): void
```
mnapoli, line 18: Can you move this to `DI\Test\IntegrationTest\Definitions\CreateDefinitionTest` and change the signature like the other test methods to merge both tests in 1:

```suggestion
    /**
     * @dataProvider provideContainer
     */
    public function test_with_variadics(ContainerBuilder $builder)
```

tests/IntegrationTest/Issues/Issue881/ClassWithVariadicParameter.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DI\Test\IntegrationTest\Issues\Issue881;
+
+final class ClassWithVariadicParameter
```
mnapoli, line 7: This class can be moved to `DI\Test\IntegrationTest\Definitions\ObjectDefinition`

.php-cs-fixer.php
serge-kvashnin: Yes, this change disables the [PhpCsFixer](https://cs.symfony.com/doc/rules/language_construct/nullable_type_declaration.html) `nullable_type_declaration` check, preventing the master build from failing. [Since PHP 8.4](https://wiki.php.net/rfc/deprecate-implicitly-nullable-types) will deprecate this kind of declaration, I suggest creating a separate Issue/PR to address it. You can then decide whether to standardize on union types or the question mark operator (whichever you think is preferable).

tests/IntegrationTest/Issues/Issue881/ClassWithVariadicParameter.php
serge-kvashnin: Done [8854fe6](https://github.com/PHP-DI/PHP-DI/pull/883/commits/8854fe64da554a36d93b3a79cdf5045745fce58b).

tests/IntegrationTest/Issues/Issue881Test.php
serge-kvashnin: Done [8854fe6](https://github.com/PHP-DI/PHP-DI/pull/883/commits/8854fe64da554a36d93b3a79cdf5045745fce58b).

serge-kvashnin:
@mnapoli Please don't merge it yet. After digging into issues https://github.com/PHP-DI/PHP-DI/issues/619 and https://github.com/PHP-DI/PHP-DI/issues/802, and considering [Ben's comment](https://github.com/PHP-DI/PHP-DI/issues/802#issuecomment-1478713488) I added some edge case tests. Turns out, there are still a few things I need to work through. I'll dive deeper and get this sorted as soon as I can.

serge-kvashnin:
Summary of changes:

1. I introduced `ksort` because for definitions where specific parameters are defined and not ordered, the array might look like `[1 => 'bar', 0 => 'foo']`. This ensures array_slice will properly fetch the rest of the variadic parameters.
2. I also made `isVariadic` check the first priority check so that it works properly when variadic parameters are not passed.
3. I also enhanced the `ReflectionBasedAutowiring::autowire()` method with iteration through defined method injections. This involves completing method injections with parameters taken from reflection but not defined explicitly (was already done for constructor). It has nothing to do with variadic parameters, but I spotted this issue when I was defining only parameter `b` while skipping `a` (which was type-hinted).

joachim-n:
I can't comment on the code in general as I don't know this codebase at all, but I can confirm that this PR fixes #881 for me.

PR #884, opened 2024-06-01
Improve CI settings

takaram:
This PR includes some improvements about CI settings.

- Use the latest version of `actions/checkout` and `actions/cache`
- Remove PHP8.1-specific step on composer install
- Paslm has output format setting for GtiHub Actions, so use it to remove `cs2pr`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,11 +38,7 @@ jobs:
                     key: php-${{ matrix.php }}-composer-locked-${{ hashFiles('composer.lock') }}
                     restore-keys: php-${{ matrix.php }}-composer-locked-
             -   name: Install PHP dependencies
-                if: matrix.php != '8.1'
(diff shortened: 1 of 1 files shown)


mnapoli:
> Remove PHP8.1-specific step on composer install

Can you explain why?

takaram:
@mnapoli
The step seems to have been there because `ocramius/proxy-manager` was unable to be installed on PHP 8.1.
See https://github.com/PHP-DI/PHP-DI/pull/808#discussion_r829376599

Now `ocramius/proxy-manager` was removed from composer.json in https://github.com/PHP-DI/PHP-DI/commit/d2b57ecaa92ba35e4029913acce09609693efcec, and we see no problems without `--ignore-platform-req`, so let's make the job simpler.

mnapoli:
ahh ok thanks!

PR #885, opened 2024-06-11
Fix documention for php definitions

N-M:
On this [page](https://php-di.org/doc/php-definitions.html#setting-definitions-in-the-container-directl)  there is a mention of 

```
$builder->setDefinitionCache(new ApcCache());
```

setDefinitionCache function does not exist in ContainerBuilder. I replaced it with enableCompilation


mnapoli:
Thanks!

PR #886, opened 2024-06-25
Show circular list of depedencies on circular dependancy error message

bdsl:
Sometimes it can take a while to trace through the dependencies to find a circle. But the container knows the dependencies its trying to resolve when it throws this error, so outputting that in the exception message should help.

May need a bit more work to only show the circle instead of the full list of dependencies being resolved.

### DIFF
--- a/src/Container.php
+++ b/src/Container.php
@@ -344,7 +344,8 @@ class Container implements ContainerInterface, FactoryInterface, InvokerInterfac
 
         // Check if we are already getting this entry -> circular dependency
         if (isset($this->entriesBeingResolved[$entryName])) {
-            throw new DependencyException("Circular dependency detected while trying to resolve entry '$entryName'");
+            $entryList = implode(" -> ", [...array_keys($this->entriesBeingResolved), $entryName]);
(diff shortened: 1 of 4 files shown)


src/Container.php
```diff
@@ -344,7 +344,8 @@ private function resolveDefinition(Definition $definition, array $parameters = [
 
         // Check if we are already getting this entry -> circular dependency
         if (isset($this->entriesBeingResolved[$entryName])) {
-            throw new DependencyException("Circular dependency detected while trying to resolve entry '$entryName'");
+            $entryList = implode(" -> ", [...array_keys($this->entriesBeingResolved), $entryName]);
+            throw new DependencyException("Circular dependency detected while trying to resolve entry '$entryName': Dependencies: " . $entries);
```
bdsl, line 348:
```suggestion
            throw new DependencyException("Circular dependency detected while trying to resolve entry '$entryName': Dependencies: " . $entryList);
```

mnapoli:
Sounds great, can you add a new test to cover this?

bdsl:
Hi @mnapoli! I'm sorry I only just saw your message today. I've edited the existing test assertions to assert on the full new error message. Apologies for the delay.

mnapoli:
Thank you!

PR #889, opened 2024-07-13
Prepare for PHP 8.4

staudenmeir:
PHP 8.4 deprecates implicitly nullable parameter types:
https://wiki.php.net/rfc/deprecate-implicitly-nullable-types

The CS fixer rule `nullable_type_declaration_for_default_null_value` can also be removed, as `true` is its default value.

## Tests

I tried adding `8.4` to the list of tested PHP versions in `ci.yml`, but the package requirements can't be resolved yet:

Bildschirmfoto 2024-07-14 um 00 38 45

### DIFF
--- a/.php-cs-fixer.php
+++ b/.php-cs-fixer.php
@@ -57,7 +57,7 @@ return $config->setRules([
     'php_unit_method_casing' => false,
     'blank_line_between_import_groups' => false,
     'global_namespace_import' => false,
-    'nullable_type_declaration_for_default_null_value' => false,
+    'nullable_type_declaration_for_default_null_value' => true,
 ])
     ->setRiskyAllowed(true)
     ->setFinder($finder);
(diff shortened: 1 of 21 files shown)


mnapoli:
I liked the `?ObjectDefinition` notation, are we able to configure the CS to use that?

Also I'm not sure why the build fails now 😕

staudenmeir:
> I liked the ?ObjectDefinition notation, are we able to configure the CS to use that?

How do you mean? What does that look like in comparison to `?ObjectDefinition $definition = null`?

> Also I'm not sure why the build fails now 😕

The CS fixer still fails because it no longer accepts return types like `string|null` and wants to replace them with `?string`. There must have been a change in the default rules. This also affects other PRs like #886.

For this PR, I originally just ran `php-cs-fixer fix` and that also changed the return types. I reverted these changes in the second commit because that's a different topic, IMO. I can of course restore the return type changes to fix the CS build.

mnapoli:
OK let's merge this for now 👍

PR #890, opened 2024-07-13
Remove deprecated Composer option

staudenmeir:
The `--no-suggest` option is deprecated in Composer 2:

> You are using the deprecated option "--no-suggest". It has no effect and will break in Composer 3.


mnapoli:
Thanks!

PR #897, opened 2024-12-11
Update composer and tests to PHP 8.4

phpfui:
I updated to the latest dependencies.  Also up updated for PHP 8.4 nullable parameter depreciation fix.

The unit tests on the master branch had one failure originally and I did not fix that.

Also upgrading PHPUnit removed the withConsecutive() method.  I used Rector to fix, but that introduced another failure.  Since I don't know (or use directly) this library, I am not sure how to fix this test or what you are actually trying to test. So you should take a look, probably an easy fix if you know what you want.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,10 +27,10 @@
         "php": ">=8.0",
         "psr/container": "^1.1 || ^2.0",
         "php-di/invoker": "^2.0",
-        "laravel/serializable-closure": "^1.0"
+        "laravel/serializable-closure": "^2.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^9.5",
+        "phpunit/phpunit": ">=9.5",
         "mnapoli/phpunit-easymock": "^1.3",
         "friendsofphp/proxy-manager-lts": "^1",
         "friendsofphp/php-cs-fixer": "^3",
--- a/tests/IntegrationTest/Definitions/AttributeTest.php
+++ b/tests/IntegrationTest/Definitions/AttributeTest.php
@@ -138,11 +138,11 @@ class ConstructorInjection
 {
     public $value;
     public string $scalarValue;
-    public stdClass $typedValue;
-    public ?stdClass $typedOptionalValue;
+    public \stdClass $typedValue;
+    public \stdClass $typedOptionalValue;
     /** @var stdClass&\ProxyManager\Proxy\LazyLoadingInterface */
     public $lazyService;
-    public stdClass $attribute;
+    public \stdClass $attribute;
     public string $optionalValue;
 
     #[Inject(['value' => 'foo', 'scalarValue' => 'foo', 'lazyService' => 'lazyService'])]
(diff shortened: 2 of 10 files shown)


composer.json
```diff
@@ -27,10 +27,10 @@
         "php": ">=8.0",
         "psr/container": "^1.1 || ^2.0",
         "php-di/invoker": "^2.0",
-        "laravel/serializable-closure": "^1.0"
+        "laravel/serializable-closure": "^2.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^9.5",
+        "phpunit/phpunit": ">=9.5",
```
mnapoli, line 33: We need to use `^`

tests/IntegrationTest/Definitions/AttributeTest.php
```diff
@@ -138,22 +138,22 @@ class ConstructorInjection
 {
     public $value;
     public string $scalarValue;
-    public stdClass $typedValue;
-    public ?stdClass $typedOptionalValue;
+    public \stdClass $typedValue;
```
mnapoli, line 141: No code style changes please.

mnapoli (commented):
The build is failing.

tests/IntegrationTest/Definitions/AttributeTest.php
phpfui: Not sure what you mean here.  \stdClass to just stdClass?  \stdClass is sometimes uses, sometimes not in master. Best practice is FQN, so \stdClass everywhere, but I only added where I changed, often to match parameters.

But happy to change.

composer.json
phpfui: I can specify each version if you want, but it is good to be on the latest PHPUnit

tests/IntegrationTest/Definitions/AttributeTest.php
mnapoli: Let's not change lines unrelated to the PR.

Also I wouldn't call this best practice as they add unnecessary noise to the code. The performance benefit makes sense in very specific cases (and a test is not one of those).

composer.json
mnapoli: Unbound versions mean the build is not reproducible and may break as soon as there's a new major version of the dependency.

phpfui:
Finally getting back to this.

So the current master branch has the following PHPUnit error on all supported versions of PHP:
```
There was 1 failure:

1) DI\Test\IntegrationTest\ContainerDebugTest::testEntriesDefinitions
Failed asserting that 'Object (\r\n
    class = DI\Container\r\n
    lazy = false\r\n
    __construct(\r\n
        $definitions = (default value) array (\n
)\r\n
        $proxyFactory = (default value) NULL\r\n
        $wrapperContainer = (default value) NULL\r\n
    )\r\n
)' matches PCRE pattern "/^Object \(\n {4}class = DI\\Container\n/".

C:\websites\PHP-DI\tests\IntegrationTest\ContainerDebugTest.php:80
```

No idea how to fix this.

Also vimeo/psalm does not yet support 8.4, so that causes PHPUnit only support V9. But PHP 8.4 lists all sorts of depreciation issues, which is why I originally opened this PR.  There is an open PR [vimeo/psalm](https://github.com/vimeo/psalm/issues/11107) for this.

Not sure how you want to proceed.

andrewnicols:
Is there anything we can do to help move this one along so that we can support PHP 8.4?

Cheers

mnapoli:
@andrewnicols yes there is, you can pick up this PR and open a new one, the goal is to get the build green

phpfui:
This is no longer needed as @andrewnicols submitted an updated PR.  Closing.

Issue #896, opened 2024-12-09
Update composer.json to laravel/serializable-closure ^2.0


phpfui:
This package is preventing laravel/serializable-closure from updating to laravel/serializable-closure V2

Happy to submit a PR if you want, just always ask before I submit since some projects appear to be dead.

```
Transitive dependencies not required in composer.json:
laravel/serializable-closure 1.3.7 ~ 2.0.0 Laravel Serializable Closure provides an easy and secure way to serializ...

>composer why laravel/serializable-closure
php-di/php-di 7.0.7 requires laravel/serializable-closure (^1.0) 
```

mnapoli:
Yes a PR sounds great, thanks!

arokettu:
Would be nice because of https://github.com/laravel/serializable-closure/issues/103

esetnik:
@mnapoli do you know if php-di will need code changes to support `laravel/serializable-closure^2.0` or is it sufficient to publish a new version with updated composer deps?

mnapoli:
@esetnik I think code changes will be required because we are locked to 1.x at the moment: https://github.com/PHP-DI/PHP-DI/blob/master/composer.json#L30

PR #899, opened 2025-01-28
Update PHP-DI to support PHP 8.4

andrewnicols:
PHP 8.4 came out in November.

This commit originally built upon #897 but after further investigation and rebasing, I discovered that the approach in 897 covered more than is necessary so I've created a new approach.

#897 modified the composer.json to use the latest compatible version of PHPUnit (`>=9.5`) when it is perfectly safe and stable to pin to 9.6 (`^9.6`) which supports PHP 7.3 through to PHP 8.4.

As a result the changes to stop using `willReturnConsecutive()` are no longer required (though they will be in the future in some form), and there is no need to make many of the other changes.

This changset instead:
* modifies the phpunit dependency to `^9.6`
* modifies the `laravel/serializable-closure` to `^1.0 || ^2.0`
* add a build on PHP 8.4

I believe this is all that is required to make this compatible with PHP 8.0 to PHP 8.4.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
         timeout-minutes: 15
         strategy:
             matrix:
-                php: [ '8.0', '8.1', '8.2', '8.3' ]
+                php: [ '8.0', '8.1', '8.2', '8.3', '8.4' ]
                 dependency-version: [ '' ]
                 include:
                     -   php: '8.0'
--- a/composer.json
+++ b/composer.json
@@ -27,10 +27,10 @@
         "php": ">=8.0",
         "psr/container": "^1.1 || ^2.0",
         "php-di/invoker": "^2.0",
-        "laravel/serializable-closure": "^1.0"
+        "laravel/serializable-closure": "^1.0 || ^2.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^9.5",
+        "phpunit/phpunit": "^9.6",
         "mnapoli/phpunit-easymock": "^1.3",
         "friendsofphp/proxy-manager-lts": "^1",
         "friendsofphp/php-cs-fixer": "^3",


phpfui:
@andrewnicols glad to see you stepped up on this PR.  I ended up going with a different package which eliminated my need for this fix. I will close the original PR now.

williamdes (approved):
Thanks, I will import this diff into Debian then
I originally did think there was more needed for 8.4

mnapoli:
Awesome thank you! The coding style warnings are unrelated so let's not be blocked by those.

williamdes:
I am working on Debian packaging the last version and `test_closure_containing_class_name_not_fully_qualified` should be marked as not compatible with PHP 8.4

```
1) DI\Test\IntegrationTest\Definitions\FactoryDefinitionTest::test_closure_containing_class_name_not_fully_qualified with data set "compiled" (DI\ContainerBuilder Object (...))
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'DI\Test\IntegrationTest\Definitions\FactoryDefinitionTestClass'
+'FactoryDefinitionTestClass'
```

Fix it is you can

mnapoli:
Weird, the tests passed in CI for PHP 8.4, any idea?

williamdes:
> Weird, the tests passed in CI for PHP 8.4, any idea?

First idea goes to laravel serializable closure still v1 in Debian

phpfui:
The test is probably wrong. You should be able to get FQN from all versions of PHP.

I would submit a PR, but away skiing.

esetnik:
I think this resolves https://github.com/PHP-DI/PHP-DI/issues/896 as well

andrewnicols:
Yup @esetnik , it does.

andrewnicols:
> > Weird, the tests passed in CI for PHP 8.4, any idea?
> 
> First idea goes to laravel serializable closure still v1 in Debian

Laravel/Serializable-closure @ v1 is not compatible with PHP 8.4 so how is Debian handling that?

williamdes:
> > > Weird, the tests passed in CI for PHP 8.4, any idea?
> > 
> > 
> > First idea goes to laravel serializable closure still v1 in Debian
> 
> Laravel/Serializable-closure @ v1 is not compatible with PHP 8.4 so how is Debian handling that?

I was about to say: patches!
But there is no patch
And this package does not have autopackage tests: https://sources.debian.org/src/php-laravel-serializable-closure/1.3.0-1/debian/

I guess nobody did catch this, I need to run the code with a basic test and report the failure.
Let's see
Most probably it will end up in an upgrade, or backport of the compatibility fixes with patches

andrewnicols:
I don'tthink that 1.3.x is actually incompatible, just that compatibility was not declared and it is not tested.

PR #915, opened 2025-06-06
Make it possible to extend container

alleknalle:
Make it possible to extend container by auto-registering ``static::class`` instead of ``self::class``


mnapoli:
I don't think we want this (the tests fail which is also an indication): this breaks the contract that there's an entry for `DI\Container`.

If we register under a different key, then the original is not guaranteed to exist.

I think a simple workaround is to `set()` the entry in a subclass (and set it to whatever key you prefer).

PR #917, opened 2025-07-21
Add support for native PHP 8.4+ lazy proxies.

BusterNeece:
PHP 8.4 introduced support for language-native lazy proxy creation, and as a result, Ocramius's ProxyManager is deprecated for users of PHP 8.4 and above in favor of these native lazy proxies.

This PR adds an option to the `ContainerBuilder` called `useNativeLazyObjects`. The naming convention is taken from the Doctrine ORM's migration to native PHP lazy objects.

The existing Ocramius `ProxyFactory` is still intact and unchanged for users opting not to use the new functionality. Its class has been incorporated into a `ProxyFactoryInterface` that exposes common methods between it and the new `NativeProxyFactory` class. As some portions of the callable that was sent to the ProxyFactory were specific to Ocramius' proxy manager, these portions have been pulled up to that specific class in favor of a generic object builder callable that works with both services.

New integration and unit tests have also been created to ensure the functionality of the new code.

The only BC break expected from this change is that the `Container` class now expects a `ProxyFactoryInterface` instead of `ProxyFactory` itself, which will affect any custom wrapping container classes. Otherwise, no other BC breaks are expected.


src/Compiler/ObjectCreationCompiler.php
```diff
@@ -142,10 +142,8 @@ private function compileLazyDefinition(ObjectDefinition $definition) : string
         return <<<STR
                     \$object = \$this->proxyFactory->createProxy(
                         '{$definition->getClassName()}',
-                        function (&\$wrappedObject, \$proxy, \$method, \$params, &\$initializer) {
-                            \$wrappedObject = $subDefinition;
-                            \$initializer = null; // turning off further lazy initialization
-                            return true;
+                        function () {
+                            return $subDefinition;
```
mnapoli, line 146: Could this simply be `$subDefinition` instead of the closure?

src/ContainerBuilder.php
```diff
@@ -252,6 +262,30 @@ public function writeProxiesToFile(bool $writeToFile, ?string $proxyDirectory =
...
     }
 
+    /**
+     * Use PHP 8.4+'s native lazy object generation for lazy-loading proxies.
+     *
+     * @see https://php-di.org/doc/lazy-injection.html
+     *
+     * @param bool $useNativeLazyObjects If true, prefer native lazy objects if available.
+     */
+    public function useNativeLazyObjects(bool $useNativeLazyObjects) : self
```
mnapoli: I'm not sure why we need this configuration method?

src/Definition/Resolver/DefinitionResolver.php
```diff
@@ -35,8 +35,8 @@ public function resolve(Definition $definition, array $parameters = []) : mixed;
      * Check if a definition can be resolved.
      *
      * @param Definition $definition Object that defines how the value should be obtained.
-     * @psalm-param T $definition
      * @param array      $parameters Optional parameters to use to build the entry.
+     * @psalm-param T $definition
```
mnapoli: Can you revert these small changes to keep the diff simpler?

src/Proxy/NativeProxyFactory.php
```diff
@@ -0,0 +1,42 @@
...
+            throw new LogicException('Lazy loading proxies require PHP 8.4 or higher.');
+        }
+
+        $reflector = new \ReflectionClass($className);
+
+        return $reflector->newLazyProxy(
+            function () use ($createFunction) {
+                return $createFunction();
+            }
+        );
```
mnapoli: Could this simply be:

```suggestion
        return $reflector->newLazyProxy($createFunction);
```

?

tests/IntegrationTest/Fixtures/LazyDependency.php
```diff
@@ -12,8 +12,10 @@
 #[Injectable(lazy: true)]
 class LazyDependency
 {
+    private bool $value = true;
+
     public function getValue(): bool
     {
-        return true;
+        return $this->value;
```
mnapoli, line 19: Why this change?

tests/IntegrationTest/NativeProxyTest.php
```diff
@@ -0,0 +1,87 @@
...
+ * @requires PHP 8.4
+ */
+class NativeProxyTest extends \DI\Test\IntegrationTest\BaseContainerTest
+{
+    /**
+     * @test
+     * @dataProvider provideContainer
+     */
+    #[\PHPUnit\Framework\Attributes\Test]
+    #[\PHPUnit\Framework\Attributes\DataProvider('provideContainer')]
```
mnapoli: Can you import the attribute classes (e.g. `\PHPUnit\Framework\Attributes\Test`) to shorten this? (this is purely cosmetic)

mnapoli (commented):
Thank you for working on this, this is awesome!

Can you confirm that if I'm on PHP 8.4 and I use lazy loading, then it's going to use the native proxies automatically?

Could you update https://github.com/PHP-DI/PHP-DI/blob/master/doc/lazy-injection.md as well? (no need for a full rewrite, just making sure it's correct).

tests/IntegrationTest/Fixtures/LazyDependency.php
BusterNeece: @mnapoli In my testing, it turns out PHP's actually kinda clever about how it generates lazy proxies...if it detects that a class has no internal "state", it won't generate a proxy and will instead just return the thing as-is. Thus, for tests to properly respond to this class, it has to have some sort of "state" to it, so PHP actually makes it lazy.

tests/IntegrationTest/NativeProxyTest.php
BusterNeece: @mnapoli Certainly, this was just in keeping with the styling of other tests in the same folder.

mnapoli: ahh I know the feeling, sorry you're caught up in-between styles of different contributions 😅

tests/IntegrationTest/Fixtures/LazyDependency.php
mnapoli: wow awesome, thanks!

src/ContainerBuilder.php
BusterNeece: @mnapoli I wasn't sure if you'd want to go with the explicit declarative approach or to just enable native proxies for anyone on 8.4 or newer. I'm happy to switch the logic to use the latter.

src/Compiler/ObjectCreationCompiler.php
BusterNeece: @mnapoli I think in this case (as opposed to the other, where it can be), it can't, because `$subDefinition` is something like `$this->subEntry123()`, and as far as I know you can't pass a function call like that directly as a callable.

BusterNeece:
Looks like if I switch the logic so that anyone on 8.4 or newer uses native proxies, then that breaks a fair number of the tests because they're expecting the specific return type and signature of the Ocramius implementation. I'll work on getting those tests updated. May be a little more work than expected if we go that route.

src/ContainerBuilder.php
mnapoli: If we can avoid the configuration and make things "just work" for users that's a huge win! It also means that once we drop support for 8.3 we no longer have to deal with ProxyManager

BusterNeece: @mnapoli Agreed. Tests should be good to re-run now!

BusterNeece:
I updated the code to automatically enable native lazy objects for anyone using PHP 8.4 or newer, and I've updated the tests and docs accordingly.

BusterNeece:
Looks like the last thing keeping it from being all green is that minor phpdoc positioning thing in a file I didn't modify (well, I did but I reverted it).

mnapoli:
ah I see, nevermind about that then (CS rule changes is annoying 😢), I need to take some time to review 👍

mnapoli:
Thank you!

PR #919, opened 2025-08-15
Fix for 7.1 after #917

### LINKED ISSUE #918: Fatal error after updating to 7.1 from 7.0.11
```
PHP Fatal error:  Could not check compatibility between DI\Proxy\ProxyFactory::createProxy(string $className, Closure $createFunction): ProxyManager\Proxy\LazyLoadingInterface and DI\Proxy\ProxyFactoryInterface::createProxy(string $className, Closure $createFunction): object, because class ProxyManager\Proxy\LazyLoadingInterface is not available in {REDACTED}/vendor/php-di/php-di/src/Proxy/ProxyFactory.php on line 42
```
It seems to be that there are missing dependencies in composer.json?

EDIT: removed local path from error

vhd:
Should fix #918 after #917

### DIFF
--- a/src/Proxy/ProxyFactory.php
+++ b/src/Proxy/ProxyFactory.php
@@ -9,7 +9,6 @@ use ProxyManager\Factory\LazyLoadingValueHolderFactory;
 use ProxyManager\FileLocator\FileLocator;
 use ProxyManager\GeneratorStrategy\EvaluatingGeneratorStrategy;
 use ProxyManager\GeneratorStrategy\FileWriterGeneratorStrategy;
-use ProxyManager\Proxy\LazyLoadingInterface;
 
 /**
  * Creates proxy classes.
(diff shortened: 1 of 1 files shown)


vhd:
PoC script just in case
```php
<?php

declare(strict_types=1);

interface SomeInterface
{
    public function someMethod(): object;
}

class SomeBrokenClass implements SomeInterface
{
    public function someMethod(): NonExistentClass { }
}

class SomeWorkingClass implements SomeInterface
{
    public function someMethod(): DateTimeImmutable { }
}

```
Should produce 
```
PHP Fatal error:  Could not check compatibility between SomeBrokenClass::someMethod(): NonExistentClass and SomeInterface::someMethod(): object, because class NonExistentClass is not available
```

It passed tests just because of those lines
```php
$proxyFactory = (\PHP_VERSION_ID >= 80400) ? new NativeProxyFactory : new ProxyFactory;
```

mnapoli:
Thank you!

